PR #2, opened 2022-09-08
implement bone lines thickness slider in overlay config

Emyka:
Just a simple config to change bone thickness, slider float from 0.01 to 15


chirpcodes:
Looks good!

PR #3, opened 2022-09-09
implement bone categories

Emyka:
If anyone can review this code, i accept critics.
I'm worried my lack of knowledge in C# created bloats or ugly coding.

Especially at those small functions
```cs
if (bone.ShouldBeDrawn(Plugin)) {
```
The Bone class doesn't have Ktisis  plugin object (to get configurations), and I didn't want to add it to the constructor, so the quick solution i found was to pass it as parameter.


Fayti1703:
Can you rebase your pull request to remove the merge commits?
(You could also fixup the cleanup commits into the main commit and clean up the commit message formatting during that rebase if you do it `-i`nteractively)

Emyka:
rebase done,  right?

Emyka:
not quite, sorry, rebase beginner conflict hell here

Emyka:
finally done 🥵

Fayti1703 (changes requested):
One main point I want to pull from the review separately, since it would be a large architectural change:
You currently keep track of categories via an integer ID -- it's probably worth moving that into a separate class, with name and default color as fields or properties.

Ktisis/Overlay/SkeletonEditor.cs
```diff
@@ -175,18 +175,21 @@ public sealed class SkeletonEditor {
 					if (bone.IsRoot)
 						continue;
 
+					var boneColor = 0x90ffffff;
+					if (bone.HkaBone.Name != null) boneColor = bone.CategoryColor(Plugin);
```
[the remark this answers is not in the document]
Emyka: I guess such a convert function could be used somewhere else, so maybe I should put it in a Util class? with both directions

Ktisis/Structs/Bones/BoneCategories.cs
```diff
@@ -0,0 +1,221 @@
...
+		{
+			BonesCategornies.TryGetValue(boneName, out int catNumber);
+			//if (catNumber == null) throw new Exception($"Unable to guess the category for bone {boneName}");
+			return catNumber;
+		}
+		public static string GetCategoryName(string boneName)
+		{
+			string? categoryName;
+			_ = Categories.TryGetValue(GetCategory(boneName), out categoryName);
+			if (categoryName == null) throw new Exception($"Unable to guess the category name for bone {boneName}");
```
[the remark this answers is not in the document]
Emyka: Ah yes, I was being pushed by the intellisense, but I figured making it nullable friendly would be better
I'll do that

Ktisis/Structs/Bones/BoneCategories.cs
```diff
@@ -0,0 +1,221 @@
...
+			{ 1, new Vector4(1.0F, 0.0F, 0.0F, 0.5647059F) },
+			{ 2, new Vector4(0.0F, 1.0F, 0.0F, 0.5647059F) },
+			{ 3, new Vector4(0.0F, 0.0F, 1.0F, 0.5647059F) },
+			{ 4, new Vector4(1.0F, 1.0F, 0.0F, 0.5647059F) },
+			{ 5, new Vector4(1.0F, 0.0F, 1.0F, 0.5647059F) },
+			{ 6, new Vector4(0.0F, 1.0F, 1.0F, 0.5647059F) },
+			{ 7, new Vector4(1.0F, 1.0F, 1.0F, 0.5647059F) },
+			{ 8, new Vector4(1.0F, 1.0F, 1.0F, 0.5647059F) }, // animal ears
+		};
+		private static readonly Dictionary<string, int> BonesCategornies = new Dictionary<string, int>() {
```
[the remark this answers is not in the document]
Emyka: whoops, sorry ^^

Ktisis/Structs/Bones/Bone.cs
```diff
@@ -112,5 +114,32 @@ public class Bone {
...
+			if (this.HkaBone.Name == null) throw new Exception("bone name must not be null");
+			return BoneCategories.GetCategory(this.HkaBone.Name);
+
+		}
+		public string CategoryName()
+		{
+			if (this.HkaBone.Name == null) throw new Exception("bone name must not be null");
+			return BoneCategories.GetCategoryName(this.HkaBone.Name);
+		}
+		public UInt32 CategoryColor(Ktisis plugin)
```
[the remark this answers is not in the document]
Emyka: Ah yes! that's what I was missing, thanks!

Ktisis/Overlay/SkeletonEditor.cs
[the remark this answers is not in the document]
chirpcodes: Yeah, convert functions would fit into the MathHelpers class

Emyka:
Thank you for the review! 😊 it gave me a good idea of better practice!

I finished fixing the mistakes and ulgies
- methods `Category()` and `CategoryName()` into a property
- `public UInt32 CategoryColor(Bone Bone)` and `public bool IsBoneVisible(Bone bone)` are now in `Configuration` class
- a bone with a null name or unable to find the category will fall in the default category, so there is no more null to handle further downstream, thus the exception throws have been removed. It also simplified some locations where it was assigning defaults

Now I am thinking how arrange the Category class.
Does it still goes into the `Ktisis.Structs.Bones` namespace?
Also, I thought `Ktisis.Data` namespace could be a good location for all the data with bone names and category name.

Should I leave it in 2 commits or squash them into one?

chirpcodes:
> Now I am thinking how arrange the Category class. Does it still goes into the `Ktisis.Structs.Bones` namespace? Also, I thought `Ktisis.Data` namespace could be a good location for all the data with bone names and category name.

`Ktisis.Structs.Bones` works fine. `Ktisis.Data` is mainly reserved for game data that needs to be parsed using Lumina - will rename it to `GameData` to make the distinction more clear.

> Should I leave it in 2 commits or squash them into one?

Up to you! No problem either way.

Fayti1703:
> Should I leave it in 2 commits or squash them into one?

Would make more sense to keep the history here (especially since we're discussing it!), so leave it as two separate commits.

Emyka:
I have a problem, in some places, I'm using the new Category as array key.
It's working fine until it's attempting to read the json configuration file.
I'm trying to find a way so it uses the category name instead when serializing and deserializing, but I can't find info...

It looks like we are using `Newtonsoft.Json`  do you know anything on this?

Fayti1703:
We're indeed using `Newtonsoft.Json` (indirectly via Dalamud). If you want more general info, you can check (their documentation)[https://www.newtonsoft.com/json/help/html/Introduction.htm], though from what I can see the functionality not *that* well documented.

I would personally recommend adding a `Newtonsoft.Json.JsonConverterAttribute` (via `[JsonConverter(typeof(ConverterClass))]`; see [their usage samples](https://www.newtonsoft.com/json/help/html/JsonConverterAttributeProperty.htm)) to the properties/fields in question, and writing your own [JsonConverter](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonConverter_1.htm) implementation.

(Also it sounds like you have quite a bit of uncommitted work; it might be worth committing and pushing what you have so far.)

Emyka:
Alright, I'll push but without the JsonConverter implementation, it's unable to read configs.

Ktisis/Configuration.cs
```diff
@@ -21,6 +23,26 @@ public class Configuration : IPluginConfiguration {
 
 		public bool DrawLinesOnSkeleton { get; set; } = true;
 		public float SkeletonLineThickness { get; set; } = 2.0F;
+        public UInt32 CategoryColor(Bone bone)
```
Fayti1703, line 26: Indent slip here.

Ktisis/Structs/Bones/BoneCategories.cs
```diff
@@ -0,0 +1,212 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Numerics;
+
+namespace Ktisis.Structs.Bones {
+	public class BoneCategories {
+
+		public static readonly List<Category> Categories = new() {
```
Fayti1703, line 9: Suggest moving this to the `Category` class itself.
Might also be worth converting this back into a dictionary (with parameters `<string, Category>`) for easier lookup.

Ktisis/Util/MathHelpers.cs
```diff
@@ -92,5 +94,11 @@ public class MathHelpers {
 
 			return angle;
 		}
+
+		public static UInt32 ConvertVector4ToUInt32(Vector4 color)
```
Fayti1703, line 98: Is this method actually necessary? It looks like it can be trivially inlined at no real cost to readability.

Ktisis/Interface/ConfigGui.cs
```diff
@@ -3,7 +3,15 @@
 
 using ImGuiNET;
 
+using Dalamud.Interface.Components;
```
Fayti1703, line 6: Double `using` statements here and in a couple of other places. Looks like an issue with conflict resolution. (If you need help, ask.)

Ktisis/Interface/ConfigGui.cs
```diff
@@ -83,16 +91,77 @@ internal class ConfigGui {
...
+					}
+					Cfg.Save(Plugin);
+				}
+
+				ImGui.Text("Bone colors by category");
+
+				foreach (Category category in BoneCategories.Categories)
+				{
+					Vector4 categoryColor = new Vector4(1.0F, 1.0F, 1.0F, 0.5647059F);
+					if (Cfg.BoneCategoryColors.ContainsKey(category)) categoryColor = Cfg.BoneCategoryColors[category];
```
Fayti1703, line 155: Can you refactor this to `TryGetValue`?

Emyka:
I'm really missing something in rebase, i'm seeing codes getting duplicated... sorry about this 😓

Fayti1703 (changes requested):

Ktisis/Configuration.cs
Emyka: Ah yes, please don't review it just now, I'm going to review all my changes, making a clean clone and clean the extra folders I have. I think I restarted coding with a local that wasn't rebased, now i'm trying to clean this mess

Fayti1703: If you need help, ask. (Though preferably not in this discussion -- I'm `Fayti1703#5459` on Discord, if you use that.)

Emyka: I might take the offer ^^' i sent a request.
Actually it is not so bad, maybe I actually made those mistakes, but when i saw the double `using` (below), I thought i made another mistake.

Ktisis/Interface/ConfigGui.cs
Emyka: Actually I can't, it seems `TryGetValue` erases the default when it's unable to find the key.
So the color is `#00000000`.

Fayti1703: It also returns `false`, so you can just set it to the correct color in that case.

Emyka: Ah! it occurred to me that could be the case, but thought it wouldn't, since it goes fine in the `Vector4` type.

Emyka: i had to convert some of the dictionaries to nullable type for this, but I think it is still cleaner than before.
Also I suppose it is better for performance.

Ktisis/Configuration.cs
```diff
@@ -26,25 +26,25 @@ public class Configuration : IPluginConfiguration {
...
-			return ImGui.GetColorU32((Vector4)color);
+			return ImGui.GetColorU32(color);
 		}
 		public bool IsBoneVisible(Bone bone)
 		{
-			ShowBoneByCategory.TryGetValue(bone.Category.Name, out bool? boneVisible);
-			boneVisible ??= true;
-			return DrawLinesOnSkeleton && (bool)boneVisible;
+			if (!ShowBoneByCategory.TryGetValue(bone.Category.Name, out bool boneVisible))
+				return true;
```
Fayti1703, line 39: This returns `true` even if `DrawLinesOnSkeleton` is `false` -- is that intentional?

Emyka: yes, this is in case a bone is uncategorizable so it shows it.
If it returned false, we would never hear of uncategorizable bones.
Maybe I should make a special category for those 🤔

sprocketaudio:
Just to chime in here - is it worth considering IVCS bones for your categories?

https://docs.google.com/document/d/1yo-ynhqH-gMTN7SQv55x2yzOe_V2II9dSAJE8FnEghE/edit#heading=h.74gj9ktdzmq

Ktisis/Configuration.cs
Emyka: whoops, i take this back, no it's a mistake, it should return `DrawLinesOnSkeleton`

Emyka:
> Just to chime in here - is it worth considering IVCS bones for your categories?
> 
> https://docs.google.com/document/d/1yo-ynhqH-gMTN7SQv55x2yzOe_V2II9dSAJE8FnEghE/edit#heading=h.74gj9ktdzmq

Yes I'm aware of IVCS.
For now these bones will fall in "body", but I can add them in special categories.

One problem is the categories will show up in the menu, even if they are empty.
I should make it so it doesn't show up. That way we would only see them in the color options, where we could eventually have an option like "show experimental categories", or maybe a more global setting for experimental or IVCS.

The thing is, I don't want to show everyone there is NSFW stuff.
But hidden behind a config they activate by their own responsibility, that sounds fine.

sprocketaudio:
IMO you do not need a tick box to show IVCS if you can make the categories not show when empty? This just adds a layer or unusability for the user. While I agree not showing NSFW by default, anyone installing IVCS isnt un-aware of its NSFW nature! If you can implelemt auto hiding of empty categories, and eventually categorise IVCS this would hugely help with using IVCS (as currrently using Ana you have to just search a huge uncategorised list of bones). Then finally, a custom category for any user amde/custom bones that fall out side this (i've added custom bones to gair for myself before for example)?

Fayti1703:
A potential problem I can see with hiding empty categories is that you might want to configure a category's color or visibility status even if no bone assigned to that category is currently being rendered.

I *suppose* we could show categories permanently (well, for the process execution) as soon as we see a bone belonging to that category or if there are user settings, but that sounds a little iffy.

Of course, if there's some way to query all registered model bones, that would also solve those issues.

Emyka:
> IMO you do not need a tick box to show IVCS if you can make the categories not show when empty? This just adds a layer of unusability for the user.

Yes, I agree.

> If you can implelemt auto hiding of empty categories, and eventually categorise IVCS this would hugely help with using IVCS (as currrently using Ana you have to just search a huge uncategorised list of bones).

I have added IVCS bones to the list, added them to their respective hand categories, added feet and ivcs (fancy name for NSFW) categories.
And the "custom" category would be where new bones end up.

> A potential problem I can see with hiding empty categories is that you (as the user) might want to configure a category's color or visibility status even if no bone assigned to that category is currently being rendered.

Yes, this is why I think it's better to make all the categories visible in the `ConfigGUI`, and only hide them in the `KtisisGUI`.
We can use code names such as "ivcs" to categorize NSFW bones.


Now I think about it, the "ivcs" category I have created does not include all the IVCS bones, as some have been distributed to hands and feet, this may create ambiguity.

> I suppose we could show categories permanently (well, for the process execution) as soon as we see a bone belonging to that category or if there are user settings, but that sounds a little iffy.

I'm not sure about that, because there are categories that would bloat the GUI. Tail and ears categories would show up after they have played those races once.

> Of course, if there's some way to query all registered model bones, that would also solve those issues.
This is the wall i'm hitting right now.
To check if a category is empty, I need to register a bone to a category, but for this, I would need to put an instance of `Category` list in either `SkeletonEditor` or `Ktisis` class.
And then `BoneCategories` would need to access this instance. `BoneCategories` is called by `Bone` class, which is referenced by several classes.
Is there a way to insert and keep data without editing all these classes?

sprocketaudio:
Another option I guess might work if you are having trouble with the hiding/categorising of IVCS might be to add a single custom category 'Other' for ALL IVCS bones (and other unknown bones) which are found. This would always show, and then have a config tick box to show IVCS bones in catgegories. Ticking this box would move all IVCS bones currently showing in 'Other' into special IVCS categories (that stay hidden untilthis box is ticked) that include IVCS Hands, IVCS Feet, and the rest of the IVCS categories. 

My reason for this thinking would be there are MANY extra bones in each of these catgegories and I do not think IVCS hands should be mixed in with normal hands etc

Fayti1703:
> 
> > Of course, if there's some way to query all registered model bones, that would also solve those issues.
> 
> This is the wall i'm hitting right now. To check if a category is empty, I need to register a bone to the category they belong to, but for this, I would need to put an instance of `Category` list in either `SkeletonEditor` or `Ktisis` class.

Not sure why you would need a separate list; you already have a dict containing all `Category`s, no?

Emyka:
> Another option I guess might work if you are having trouble with the hiding/categorising of IVCS might be to add a single custom category 'Other' for ALL IVCS bones (and other unknown bones) which are found. This would always show, and then have a config tick box to show IVCS bones in catgegories. Ticking this box would move all IVCS bones currently showing in 'Other' into special IVCS categories (that stay hidden until this box is ticked) that would include IVCS Hands, IVCS Feet, and the rest of the IVCS categories.

Moving categories could be a thing, but as you said, extra tick boxes may add a degree of instability for the user.
I'll consider this solution if i'm unable to find a way to auto hide categories.

> My reason for this thinking would be there are MANY extra bones in each of these catgegories and I do not think IVCS hands should be mixed in with normal hands etc

That's a fair point.
 
> This would at least give the user the ability to see all IVCS bones easily with no config needed (as they would be under 'Other') but also allow the possibility to categorise them which would show NSFW category names if they wanted to.

You mean with extra categories such as `ivcs right hand`, `ivcs left foot`, etc?

> _Note - I cant really read the code, so no idea how feasable my suggestions are! - Just throwing out ideas from a user perspective_

No worries, all suggestions are welcome!


> > > Of course, if there's some way to query all registered model bones, that would also solve those issues.
> > 
> > 
> > This is the wall i'm hitting right now. To check if a category is empty, I need to register a bone to the category they belong to, but for this, I would need to put an instance of `Category` list in either `SkeletonEditor` or `Ktisis` class.
> 
> Not sure why you would need a separate list; you already have a dict containing all `Category`s, no?

I think the dicts are destroyed and recreated each time an action is made on them.

sprocketaudio:
> You mean with extra categories such as `ivcs right hand`, `ivcs left foot`, etc?

Yeah so by default on load you would have all found extra bones under 'Other'. This would mean the user would see and can toggle all IVCS bones without worrying about config and could start using them immediately.

The config option would sort all IVCS bones into their correct categories:

ivcs body
ivcs right hand
ivcs left hand
ivcs p***s
ivcs v****a
ivcs buttocks
ivcs right foot
ivcs left foot

Fayti1703:
Would it make sense to invert the data hierarchy here? Like, instead of having a table of `bone -> category`, make each category contain a list of what bones it contains by name.

Would eliminate the dictionary lookups to associate each bone -- and you could still build a `bone -> category` Dictionary in a static constructor.

Emyka:
Ah that's a good idea.
Maybe It could populate the categories during the construction of `Category`.

Emyka:
> > You mean with extra categories such as `ivcs right hand`, `ivcs left foot`, etc?
> 
> Yeah so by default on load you would have all found extra bones under 'Other'. This would mean the user would see and can toggle all IVCS bones without worrying about config and could start using them immediately.

I decided to show every categories that have been found, so when a user starts the plugin for the first time, they see no categories.
When they enable "toggle skeleton" and enter in GPose, it populates the categories where bones are found.

> The config option would sort all IVCS bones into their correct categories: [...]

Done. 😊 

Also, new categories are are now easily made, with just associating a bone name with a category name.
It shouldn't be too hard to allow the user to edit their own category/bone association.

Ktisis/Overlay/SkeletonEditor.cs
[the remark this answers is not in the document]
Emyka: I just realised what you mean for that one, sorry 😅

Fayti1703:
I'll be honest, I figured I would just do the data hierarchy thing (6d9e14b) but then I noticed these various smaller issues and thought I'd fix them inline instead of bothering you about it

Skip forward an hour to this blob of commits.

Fayti1703 (approved):

Emyka:
Sorry for making you finish what i started 😳 looks so pro.

chirpcodes (approved):
Looks good, will merge into the main repo now. :)

PR #4, opened 2022-09-15
Tooltip (with confirmation) helper functions

Emyka:
The main purpose of this PR is to offer the method `IconButtonHoldConfirm()`, which simply replaces a button, but it won't have any effect if a key combination is not hold when clicking on it.

This is result of a suggestion by @Fayti1703, where a confirmation before erasing color settings would be better.

By-products of this PR are the methods `IconButtonTooltip()` and `Tooltip()` and the class `GuiHelpers` which offer ways to draw Gui stuff easier, and sort Gui functions in its own place.


chirpcodes (approved):
Looks good! The helper functions are awesome, this will likely help elsewhere in the plugin UI.

PR #5, opened 2022-09-16
implemented coordinate table

Emyka:
Adds a simple coordinate 3x3 grid in KtisisGui.
image

The current purpose is debugging the coordinate to understand the rotation issue.

Problem 1:
The coordinates are directly taken from `Bone.Transform`, so I think it's related to local and not world, but i didn't confirm.

Problem 2:
I don't get why the children bone aren't transformed, the `Bone.TransformBone(Transform t, List<BoneList> skeleton)` should achieve that, as far as I understand.
```c#
selectedBone.TransformBone(delta, Plugin.SkeletonEditor.Skeleton);
```


chirpcodes:
Merged so that debugging can start asap, I believe I can address problem 2 tomorrow. Thank you!

PR #6, opened 2022-09-18
categories and thickness tweaks

Emyka:
Bundled PR with various tweaks:

1. new category 'weapons' and moved the respective bones in it
2. lines thickness changes according to zoom level
3. dot radius now changes according to line thickness configuration
4. attempt to fix color category overlap

There are issues I couldn't resolve:

- Line thickness and Dot radius math are a bit weird, it somehow works for me, but if anyone has a better idea, it's welcome
- the parenting seems to have a strange effect on the category overlap fix, as if it's jumping from child to grand-parent

### DIFF
--- a/Ktisis/Overlay/SkeletonEditor.cs
+++ b/Ktisis/Overlay/SkeletonEditor.cs
@@ -224,6 +224,7 @@ namespace Ktisis.Overlay {
 
 							Dalamud.GameGui.WorldToScreen(parentPos, out var pPos);
 							uint lineColor = boneColor;
+							if (bone.Category != parent.Category) lineColor = ImGui.GetColorU32(Ktisis.Configuration.GetCategoryColor(parent));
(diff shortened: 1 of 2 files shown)


Ktisis/Overlay/SkeletonEditor.cs
```diff
@@ -223,7 +223,12 @@ public sealed class SkeletonEditor {
 							var parentPos = model->Position + parent.Rotate(model->Rotation) * model->Height;
 
 							Dalamud.GameGui.WorldToScreen(parentPos, out var pPos);
-							draw.AddLine(pos, pPos, boneColor, Ktisis.Configuration.SkeletonLineThickness);
+							uint lineColor = boneColor;
+							if (bone.Category != parent.Category) lineColor = ImGui.GetColorU32(Ktisis.Configuration.GetCategoryColor(parent));
```
Emyka, line 227: This is where the category overlap is being checked, but the result isn't as I would have expected.
image

Fayti1703:
Rebased to split out the individual changes.

PR #10, opened 2022-10-09
Sets browse + QoL for item search lists

Emyka:
This is mostly a lot of QoL changes in `EditEquip` window.

Implements:
 - gives the possibility to browse Gear sets (Glamour plate not implemented yet).

Qol:
- keyboard focus the input search box when opening a search list (`DrawSelectorList` and `DrawSetSelectorList`).
- up/down to browse, page up/down to browse faster
- escape to close the search list
- apply the first choice on press enter after typing
- the window is scrolled automatically to follow the current selection when using keys
- the current selection is highlighted
- search input text is case insensitive

Additional info:
Created a new struct for `EquipmentSet`, this is where the sets (GearSet, Glam Plates, etc)   data collection is made.
There is a subclass `EquipmentSets` (plural) in the same file. I'm not sure it was very smart, but I didn't want to create too much files for this.
`EquipmentSets` is used as instance in `EditEquip` for various reasons.

`DrawSetSelectorList` is copy pasted from `DrawSelectorList`. I tried to combine the common code, but I figured they were quite different despite their similarities, and they could grow even more different in time.

### DIFF
--- a/Ktisis/Interface/Windows/ActorEdit/EditEquip.cs
+++ b/Ktisis/Interface/Windows/ActorEdit/EditEquip.cs
@@ -19,12 +19,6 @@ namespace Ktisis.Interface.Windows.ActorEdit {
 		public const int _IconSize = 36;
 		public static Vector2 IconSize = new(_IconSize, _IconSize);
 
-		public const ImGuiKey KeyBindBrowseUp = ImGuiKey.UpArrow;
-		public const ImGuiKey KeyBindBrowseDown = ImGuiKey.DownArrow;
-		public const ImGuiKey KeyBindBrowseUpFast = ImGuiKey.PageUp;
-		public const ImGuiKey KeyBindBrowseDownFast = ImGuiKey.PageDown;
-		public const int FastScrollLineJump = 8; // number of lines on the screen?
-
 		// Properties
 
 		public unsafe static Actor* Target => EditActor.Target;
@@ -33,12 +27,10 @@ namespace Ktisis.Interface.Windows.ActorEdit {
 
 		public static Dictionary<EquipSlot, ItemCache> Equipped = new();
 
-		public static Vector2 SelectPos;
 		public static EquipSlot? SlotSelect;
 		public static IEnumerable<Item>? SlotItems;
 		public static string ItemSearch = "";
 		public static string SetSearch = "";
-		public static int LastSelectedItemKey = 0;
 		public static bool DrawSetSelection = false;
 		public static EquipmentSets? Sets = null;
 
@@ -122,12 +114,10 @@ namespace Ktisis.Interface.Windows.ActorEdit {
 		public static void OpenSelector(EquipSlot slot) {
 			SlotSelect = slot;
 			SlotItems = Items!.Where(i => i.IsEquippable(slot));
-			SelectPos = ImGui.GetMousePos();
 		}
 
 		public static void OpenSetSelector() {
 			DrawSetSelection = true;
-			SelectPos = ImGui.GetMousePos();
 		}
 
 		public static void DrawControls()
@@ -148,8 +138,8 @@ namespace Ktisis.Interface.Windows.ActorEdit {
 			if (SlotItems == null)
 				return;
 
-			HoverPopupWindow(
-					HoverPopupWindowFlags.SelectorList | HoverPopupWindowFlags.SearchBar,
+			GuiHelpers.HoverPopupWindow(
+					GuiHelpers.HoverPopupWindowFlags.SelectorList | GuiHelpers.HoverPopupWindowFlags.SearchBar,
 					SlotItems,
 					(i) => false, // draw Before Line
 					(i) => i.Name, // lineLabel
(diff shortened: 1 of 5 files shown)


Ktisis/Interface/Windows/ActorEdit/EditEquip.cs
```diff
@@ -125,30 +144,152 @@ public class EditEquip {
...
+			if (DrawSetSelection)
+				DrawSetSelectorList();
+
+			ImGui.Separator();
+		}
+
+
+
+
+		// dirty shameless copy paste of DrawSelectorList(), merge common code?
```
Fayti1703, line 216: ^

Ktisis/Structs/Actor/EquipmentSet.cs
```diff
@@ -0,0 +1,208 @@
...
+		{
+			ItemsSheet = itemsSheet;
+			Sets = new();
+		}
+
+		// helpers
+		private Item GetEmperorNewItemForSlot(EquipIndex equipIndex) => ItemsSheet!.Where(i => i.IsEquippable(Equipment.EquipIndexToItemSlot(equipIndex)) && i.Name.Contains("Emperor's New")).First(); // TODO: improve performances, maybe cache them in constructor
+
+		// public managers
+		public static EquipmentSets Init(IEnumerable<Item>? itemsSheet) => new EquipmentSets(itemsSheet);
```
Fayti1703, line 41: This method looks unnecessary -- the constructor is already public.

Ktisis/Interface/Windows/ActorEdit/EditEquip.cs
```diff
@@ -41,9 +52,12 @@ public class EditEquip {
 		// UI Code
 
 		public unsafe static void Draw() {
+			DrawControls();
```
Fayti1703, line 55: Move this down below the initialization of `Items`…

Ktisis/Interface/Windows/ActorEdit/EditEquip.cs
```diff
@@ -125,30 +144,152 @@ public class EditEquip {
...
 					SlotSelect = null;
 					SlotItems = null;
 				}
 			}
 
+			ImGui.End();
+		}
+		public static void DrawControls()
+		{
+			Sets = EquipmentSets.Init(Items);
```
Fayti1703, line 202: And non-null assert here...
```suggestion
			Sets = new EquipmentSets(Items!);
```

Ktisis/Structs/Actor/EquipmentSet.cs
```diff
@@ -0,0 +1,208 @@
...
+			Source = source;
+		}
+	}
+
+	public class EquipmentSets
+	{
+		public List<EquipmentSet> Sets;
+		private IEnumerable<Item>? ItemsSheet;
+
+		public EquipmentSets(IEnumerable<Item>? itemsSheet)
```
Fayti1703, line 31: And then remove nullability here (and propagate to fields &c)

Fayti1703 (changes requested):

Ktisis/Interface/Windows/ActorEdit/EditEquip.cs
Emyka: Well for this, the two methods have become different, I tried making only one with dynamic parameters but I couldn't cast them or use use .Where() once inside the method. After some struggling I gave up and reverted the changes.

Also it ended up having one callback to draw the line and another one when a line is selected selected.

Fayti1703: Might look at factoring out the common code this evening, though for... reasons you'll have to wait until tomorrow for any results. (Feel free to push changes in the meanwhile, though)

Otherwise... drop the comment.

Ktisis/Interface/Windows/ActorEdit/EditEquip.cs
Emyka: I ended up changing `ItemsSheet` to non-nullable and asserting when inserting items into this variable 3f9b93b. does that look OK?

Ktisis/Structs/Actor/EquipmentSet.cs
Emyka: Ah, I didn't read this comment before fixing [the previous request](https://github.com/ktisis-tools/Ktisis/pull/10#discussion_r992092496)
😅 So I guess that's done too.

Ktisis/Interface/Windows/ActorEdit/EditEquip.cs
Emyka: It should be done with 87e4845.

Ktisis/Interface/Windows/ActorEdit/EditEquip.cs
Emyka: I'm going to try the following.
I thought of making methods like `BeginPopupHooverWindow` and `EndPopupHooverWindow` with returning the `ImGui.Begin` bool on begin, and pass a few variables (like `focus`) as `ref`.
And maybe make flags to make them either `ImGui` `ListBox`, or other kind of window.

Emyka:
I'm quite satisfied with the `HoverPopupWindow` (maybe that name isn't appropriate though).
It is now completely independent from `EditEquip`.

It's a bit complex and i'm worried it won't be so easy to implement into for future uses. I'm going to try adding flexibility on it.
Maybe 2d navigation for dye selection, width, different position method.

PR #17, opened 2022-10-30
Transform Table fix

Emyka:
This change makes the Transform Table to uses the Gizmo to apply transformations.

It is not bug-free, but as it uses the gizmo methods, it's easier to blend with the "path" chirp has made for gizmo manipulation.

### DIFF
--- a/Ktisis/Interface/Components/TransformTable.cs
+++ b/Ktisis/Interface/Components/TransformTable.cs
@@ -4,7 +4,9 @@ using ImGuiNET;
 using FFXIVClientStructs.Havok;
 
 using Ktisis.Helpers;
+using Ktisis.Overlay;
 using Ktisis.Structs;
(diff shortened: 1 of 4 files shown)


chirpcodes:
> Other than my comment, the only complaint I have is that we'd have both a method `DecomposeMatrix` and a method `Decompose` on `Gizmo`.
> 
> The new `Decompose` seems more inline with what I think you'd _normally_ do for something like this... I suppose we could make everything else use `Decompose` and then inline `DecomposeMatrix`.
> 
> @chirpcodes, thoughts on that idea?

Sounds good. We can also remove DecomposeDelta from the class.

Emyka:
> We can also remove DecomposeDelta from the class.

I agree. I tried to use the delta matrix, but it doesn't seem connected, so I made my own delta logic.

PR #20, opened 2022-11-05
workspace user friendly changes

Emyka:
A bunch of UX changes after the suggestion thread "User friendly updates to workspace UI" with the big help of Serena ❤️.

The following parts were moved from `Workspace` and `GuiHelpers` classes to their respective new classes:

- `BoneTree`
- `ControlButtons`: Gizmo Operation & extras buttons
- `AnimationControls`
- `PopupSelect`: HoverPopupWindow methods from GuiHelpers


Ktisis/Interface/Components/PopupSelect.cs
```diff
@@ -0,0 +1,180 @@
...
+		private static int ColFromKey(int key) => key % HoverPopupWindowColumns;
+		private static int KeyFromRowCol(int row, int col) => (row * HoverPopupWindowColumns) + col;
+
+		public static void HoverPopupWindow(
+				HoverPopupWindowFlags flags,
+				IEnumerable<dynamic> enumerable,
+				Func<IEnumerable<dynamic>, string, IEnumerable<dynamic>> filter,
+				Action<dynamic> header,
+				Func<dynamic, bool, (bool, bool)> drawBeforeLine, // Parameters: dynamic item, bool isActive. Returns bool isSelected, bool Focus.
+				Action<dynamic> onSelect,
```
Fayti1703, line 52: Probably shouldn't be using dynamic here and instead use generics (if possible)

Ktisis/Util/GuiHelpers.cs
```diff
@@ -11,26 +11,30 @@
 
 using FFXIVClientStructs.Havok;
 
-using Ktisis.Structs.Bones;
-
 namespace Ktisis.Util
 {
 	internal class GuiHelpers {
-		public static bool IconButtonHoldConfirm(FontAwesomeIcon icon, string tooltip, bool isHoldingKey) {
+		public static bool IconButtonHoldConfirm(FontAwesomeIcon icon, string tooltip, bool isHoldingKey, Vector2 size = default) {
```
Fayti1703, line 17: This is going to result in a size of `(0, 0)`... is that intentional?

Fayti1703 (commented):
In addition to the notes below: Commit cfef801 has...quite a long commit summary.
Try to keep your summaries short and move extra information into the rest of the commit message, separated by a blank line.

Ktisis/Util/GuiHelpers.cs
Emyka: Yes, a size of `(0, 0)` means it's going to take the space it needs.

I wanted to keep the code clean without too much ifs and it seems this works as well as `(-1, -1)`, which was my first intention.

Ktisis/Interface/Components/PopupSelect.cs
Emyka: Very good suggestion!
Done in e10daa6.

Emyka:
Sorry for the last minute change on f11b836, we had a discussion about it.
It is ready now.

PR #27, opened 2022-11-10
Configurable keybinds

Emyka:
A system to easily add new keys and key events.

For now, it features the following events:
- switch gizmo operation to translate (on key release)
- switch gizmo operation to rotate (on key release)
- switch gizmo operation to scale (on key release)
- switch gizmo operation to universal (on key release)
- toggle gizmo mode (on key release)
- toggle hide skeleton (on key state change => hide on hold, show on release)


Emyka:
Not ready yet, reasons:
Actions are still fired when typing in chat.

PR #29, opened 2022-11-11
ConfigGui.cs Localizeable

EstoyMejor:
German translation added as well, but needs work. 
+ ImgBot optimized the image, kick that out if you like.


chirpcodes:
Looks good, nice work!

PR #34, opened 2022-11-13
Custom bone offset

Emyka:
Allow the user to adjust a bone position.
only the dot, the gizmo will still appear at the original position.

This can improve the UX for facial expression.

Features a list is to visualise, drop, export and import a single or all bones

### DIFF
--- a/Ktisis/Configuration.cs
+++ b/Ktisis/Configuration.cs
@@ -88,6 +88,7 @@ namespace Ktisis {
 
 		// Data memory
 		public Dictionary<string, GlamourDresser.GlamourPlate[]?>? GlamourPlateData { get; set; } = null;
+		public Dictionary<string,Dictionary<string,Vector3>> CustomBoneOffset { get; set; } = new();
 
 		// Validate for changes in config versions.
 
--- a/Ktisis/Configuration.cs
+++ b/Ktisis/Configuration.cs
@@ -88,7 +88,7 @@ namespace Ktisis {
 
 		// Data memory
 		public Dictionary<string, GlamourDresser.GlamourPlate[]?>? GlamourPlateData { get; set; } = null;
-		public Dictionary<string,Dictionary<string,Vector3>> CustomBoneOffset { get; set; } = new();
+		public Dictionary<string, Dictionary<string, Vector3>> CustomBoneOffset { get; set; } = new();
 
 		// Validate for changes in config versions.
 
--- a/Ktisis/Configuration.cs
+++ b/Ktisis/Configuration.cs
@@ -88,7 +88,7 @@ namespace Ktisis {
 
 		// Data memory
 		public Dictionary<string, GlamourDresser.GlamourPlate[]?>? GlamourPlateData { get; set; } = null;
-		public Dictionary<string,Dictionary<string,Vector3>> CustomBoneOffset { get; set; } = new();
+		public Dictionary<CustomOffset.BodyType,Dictionary<string,Vector3>> CustomBoneOffset { get; set; } = new();
 
 		// Validate for changes in config versions.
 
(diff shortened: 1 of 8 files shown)


Ktisis/Structs/Bones/CustomOffset.cs
```diff
@@ -0,0 +1,50 @@
...
+			var gender = actor->Customize.Gender;
+			var race = actor->Customize.Race;
+
+			if (gender == Gender.Male)
+				switch (race) {
+					case Race.Lalafell: return BodyType.LalafellMale;
+					case Race.Roegadyn:
+					case Race.Hrothgar: return BodyType.HrothgarRoegadyn;
+					default: return BodyType.TallMale;
+				} else switch (race) {
```
Fayti1703, line 31: This is not correctly indented.
At minimum, I'd suggest this:
```suggestion
				} 
			else 
				switch (race) {
```

But ideally, add braces around this. It's a multi-line statement, we can afford those.

Ktisis/Util/GuiHelpers.cs
```diff
@@ -87,6 +89,19 @@ internal class GuiHelpers {
 			}
 		}
 
+		public static bool DragFloat3FillWidth(string label, bool icon, string? tooltip, ref Vector3 vector,float speed, string format) {
```
Fayti1703, line 92: Minor, but:
```suggestion
		public static bool DragFloat3FillWidth(string label, bool icon, string? tooltip, ref Vector3 vector, float speed, string format) {
```

Ktisis/Configuration.cs
```diff
@@ -88,6 +88,7 @@ public class Configuration : IPluginConfiguration {
 
 		// Data memory
 		public Dictionary<string, GlamourDresser.GlamourPlate[]?>? GlamourPlateData { get; set; } = null;
+		public Dictionary<string,Dictionary<string,Vector3>> CustomBoneOffset { get; set; } = new();
```
Fayti1703, line 91:
```suggestion
		public Dictionary<string, Dictionary<string, Vector3>> CustomBoneOffset { get; set; } = new();
```

Ktisis/Interface/Windows/ConfigGui.cs
```diff
@@ -309,7 +313,7 @@ internal static class ConfigGui {
 		}
 		private static string PrettyKeys(List<VirtualKey>  keys) => string.Join(" + ", keys.Select(k => VirtualKeyExtensions.GetFancyName(k)));
 
-		public static void DrawDataTab(Configuration cfg) {
+		public static unsafe void DrawDataTab(Configuration cfg) {
```
Fayti1703, line 316: This doesn't need to be made unsafe.

Fayti1703 (commented):
Found some minor formatting problems. Looks good otherwise!

Ktisis/Structs/Bones/CustomOffset.cs
Emyka: VS offers this, i'm not sure if it's worth.
```suggestion
			else return race switch {
				Race.Lalafell => BodyType.LalafellFemale,
				_ => BodyType.TallFemale,
			};
```

Fayti1703: I would still break the line there. Could do both?

Emyka: Yes I noticed it was unrelated with the else being on the wrong indent (fixed with 1acc264).
I was curious of your opinion about "switch statement expression" style offered by vs.

Fayti1703: Switch expressions are neat, imo. Don't see a reason not to use them.
(Though if you apply it to the `else` branch, you should probably apply it to the `if` branch as well, for consistency)

PR #35, opened 2022-11-13
Added an event manager.

therealedited:
Alongside the event manager, the event `GPoseChange` was added.

This event is fired in two cases:
- The actor entered GPose.
- The actor left GPose.

Any method that subscribes to it will receive an enum `ActorGposeState` as input that indicates the current gpose state of the actor.

### DIFF
--- a/Ktisis/Events/EventManager.cs
+++ b/Ktisis/Events/EventManager.cs
@@ -1,0 +1,18 @@
+﻿using Ktisis.Structs.Actor.State;
+
+namespace Ktisis.Events
+{
+	public static class EventManager
+	{
+		public unsafe delegate void GPoseChange(ActorGposeState state);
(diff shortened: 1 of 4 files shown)


Ktisis/Events/EventManager.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Ktisis.Structs.Actor.State;
+
+namespace Ktisis.Events {
+	public static class EventManager {
+		public unsafe delegate void GPoseChange(ActorGposeState state);
```
Fayti1703, line 5: Why is this delegate unsafe? There's no pointers or anything of the sort here.

Fayti1703 (changes requested):

Fayti1703 (approved):

PR #36, opened 2022-11-13
More translation

EstoyMejor:
PRing current progress :)


EstoyMejor:
as for the literal meaning, I was thinking about that as well, we could go for something more generic, but taht would require translators to cross reference the english file in order to translate in context. I dont know how its done in prod tho so im happy to change my pattern if you have propositions.

Fayti1703:
From what I've seen, typically you'd make the translation key something like `config.input_tab.title` -- basically, describe what the text *is* ("the title for the input tab of the configuration menu") and not what the text *says*.

Open to discussions on that front, of course -- but I personally think that's a good pattern. Especially because it avoids problems with assigning the same key in two different places because the text happens to be the same in English.

EstoyMejor:
I implemented a proposed change. Happy to discuss this as Its only a proposition :) I had the time and cobbled something together.

Fayti1703:
While that looks good for the most part (I personally think `langBones` is... maybe not the best choice there), I do also think that those changes should be in a separate PR. 
(Apologies if I made it seem like these changes would be a requirement for merging -- that was not my intention)

EstoyMejor:
Made new branch on my fork :) Will make a new PR with that later today.

Fayti1703 (approved):

PR #39, opened 2022-11-16
Add reference images

kaze-xiv:
https://trello.com/c/6rD5GpSU/25-reference-image-overlay

Adds Reference images to Ktisis. Reference images are useful for getting the posing right of a XIV character by letting the user compare their current pose to an existing visual reference.
image

Features:

* Supports multiple images.
* Hide window decoration toggle (removes titlebar and the possibility of resizing the references)
* Image transparency slider for less distraction
* Ability to toggle visibility of reference without deleting it entirely


Ktisis/Configuration.cs
```diff
@@ -44,6 +45,12 @@ public class Configuration : IPluginConfiguration {
 		public bool DrawLinesOnSkeleton { get; set; } = true;
 		public float SkeletonLineThickness { get; set; } = 2.0F;
 
+		// References
+		public Dictionary<int, ReferenceInfo> References { get; set; } = new();
```
Fayti1703, line 49: Couldn't this be a `List<ReferenceInfo>` instead? Don't see a need for a dictionary here.

Ktisis/Interface/Windows/References.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Numerics;
+
+using Dalamud.Logging;
+
+using ImGuiNET;
+using ImGuiScene;
+
+namespace Ktisis.Interface.Windows {
+	internal static class References {
+		public static Dictionary<int, TextureWrap> Textures = new();
```
Fayti1703, line 13: See other comment -- though here I have an alternate suggestion: turn this into a `string`-indexed `Dictionary` and keep a ref-count for how many `ReferenceInfo`s use a given texture. 
Dispose when deleting the reference that uses the texture.

Basically just deduplication if someone loads the same image as two references. Probably a bit overkill, but might still be nice to have.

Ktisis/Interface/Windows/ConfigGui.cs
```diff
@@ -419,5 +427,78 @@ internal static class ConfigGui {
...
+			if (GuiHelpers.IconButton(FontAwesomeIcon.Plus, ButtonSize)) {
+				cfg.References[cfg.NextReferenceKey] = new ReferenceInfo() { Showing = true };
+			}
+			ImGui.SameLine();
+			ImGui.Text(Locale.GetString("config.references_tab.add_new"));
+
+			ImGui.EndTabItem();
+		}
+
+		public static Action<bool, string> ReferenceFileSelected(Configuration cfg, int key) => (s, f) => {
```
Fayti1703, line 479: Expand the names of `s` and `f` here, please.

Fayti1703: (Also this has potential for inlining.)

Fayti1703 (changes requested):
Looks good overall, but there's a few things I'd like changed; see below.

Ktisis/Configuration.cs
kaze-xiv: I've kept `Dictionary<int, ReferenceInfo>` because (AFAIK) you need a unique ID for the imgui window. Using a key to a dictionary prevents a collision more easily than a property on the object. Using the index of the item of the list would be no good either; if you had (A, A) and deleted the first A then it would probably look like nothing happened (as for imgui, only the second one disappeared). Let me know what you think.

Ktisis/Interface/Windows/References.cs
kaze-xiv: I've opted for `DisposeUnreferencedTextures` instead of a formal refcount.

kaze-xiv:
Sorry for the delay on this one, I was a little sick.

Ktisis/Configuration.cs
Fayti1703: Uniqueness is a good point here and something I hadn't considered. I'm still not *quite* happy with this, but I wouldn't consider it a merge blocker.

Ktisis/Interface/Windows/References.cs
Fayti1703: That also works. (though it's technically less performant, for the expected data size that *should* be okay). Resolving.

Fayti1703 (approved):
Looking good.

For extra credit, you could add a comment regarding the uniqueness constraint on the window ID to the dictionary, but that's nothing merge blocking.

kaze-xiv:
Added the comment and rebased the branch just in case.

PR #44, opened 2022-11-23
Controls Buttons reposition / Multi-Selection Operation

Fia3D:
image

### DIFF
--- a/Ktisis/Util/EnumFlagExtension.cs
+++ b/Ktisis/Util/EnumFlagExtension.cs
@@ -1,0 +1,7 @@
+﻿namespace Ktisis.Util {
+	internal static class EnumFlagExtension {
+		public static T AddFlag<T>(this System.Enum type, T value) => (T)(object)((int)(object)type | (int)(object)value!);
+		public static T RemoveFlag<T>(this System.Enum type, T value) => (T)(object)((int)(object)type & ~(int)(object)value!);
+		public static T ToggleFlag<T>(this System.Enum type, T value) => (T)(object)((int)(object)type ^ (int)((object)value!));
+	}
+}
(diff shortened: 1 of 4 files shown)


Fayti1703 (changes requested):
Hate to be so immediately negative to a first-time contributor, but could you:

1. Correct the indentation on new lines
2. Remove the spurious changes on the lines just before the new lines (I'm guessing that's a CRLF/LF difference)
4. Split your changes across multiple commits (One atomic change per commit)

Fia3D:
image
Last commit allow to remove the white circle (ctrl + click on rotation)

Ktisis/Util/EnumFlagExtension.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Ktisis.Util
+{
+	static class EnumFlagExtension
+	{
+		public static bool FlagHas<T>(this System.Enum type, T value) => ((int)(object)type & (int)(object)value) == (int)(object)value;
```
Fayti1703, line 5: This already exists as an instance method on `System.Enum` -- namely `HasFlag`. The other methods here are fine, but this one in specific already has a native implementation.

If you could change your calls to that method and then remove this method here, that'd be great.

Fayti1703 (changes requested):
This is looking quite good! I would request one last change before calling this merge-ready:

Fia3D:
Should be good now

Fayti1703 (approved):

Fia3D:
Rebase to make it compatible with new merged titlebar settings

Fayti1703:
Minor thing I just noticed: Could you un-hang the braces in your code? i.e.

```diff
- if (ImGui.GetIO().KeyCtrl)
- {
+ if (ImGui.GetIO().KeyCtrl) {
```

PR #45, opened 2022-11-23
Settings buttons on title bar

Emyka:
Moved setting button near fold and close button.


Fayti1703 (approved):
Recommend "Squash and Merge" here -- don't see the history as worth preserving for this PR.

PR #46, opened 2022-11-24
Actor List tweaks

Emyka:
- Better method to find and filter objects to put in the search list
- Change (Player) for (You) when it's Actor 201
- Put non-player names instead of actors ids when "Hide character name" option is enabled.
- Allow selecting non-networked NPCs (non-targetable even outside of gpose)
- NPC or objects without a name will be displayed as Actor #xxx
- Optimize performances, clean up, comments.

### DIFF
--- a/Ktisis/Interface/Components/ActorsList.cs
+++ b/Ktisis/Interface/Components/ActorsList.cs
@@ -107,6 +107,7 @@ namespace Ktisis.Interface.Components {
 
 			var actor = (Actor*)gameObject;
 			if (actor == null) return false;
+			if (actor->Model == null) return false;
 
 			var objectKind = (ObjectKind)gameObject->ObjectKind;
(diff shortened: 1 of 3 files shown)


Fayti1703 (changes requested):
Two points on the last commit:
1. Code style and cleanup go in a separate commit (these are not part of the same atomic change since someone might need to apply the actual functional changes without applying the reformatting due to conflicts on the latter)
2. This is a really minor point, but the commit message is supposed to be in the imperative, to complete the sentence "When applied, this commit will…"

Otherwise, this looks good!

Fayti1703:
(in addition to the above, I've fixed an instance where you accidentally added a space as the first character on a line and dropped one blank line before the start of the Object-LINQ-Query for readability reasons -- at least, I think it's better if the list and LINQ-Quey are a bit closer together there.)

Fayti1703 (approved):

PR #50, opened 2022-11-27
Colored TransformTable

Fia3D:
With my skin that have really wide padding
image
With default dalamud
image

### DIFF
--- a/Ktisis/Interface/Components/TransformTable.cs
+++ b/Ktisis/Interface/Components/TransformTable.cs
@@ -64,7 +64,7 @@ namespace Ktisis.Interface.Components {
 
 			FetchConfigurations();
 
-			var axisColors = new[] {new Vector4(1, 0, 0, 1), new Vector4(0, 1, 0, 1), new Vector4(0, 0, 1, 1)};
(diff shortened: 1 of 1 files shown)


Fayti1703:
Could you factor out the common code into a method here, similar to the original `DragFloat3`?

Fia3D:
> Could you factor out the common code into a method here, similar to the original `DragFloat3`?

I added ColoredDragFloat3 and ColoredDragFloat

Ktisis/Interface/Components/TransformTable.cs
```diff
@@ -60,47 +64,73 @@ public class TransformTable {
...
 			// Scale
-			result |= ImGui.DragFloat3("##Scale", ref Scale, BaseSpeedSca * multiplier, 0, 0, DigitPrecision);
+			result |= ColoredDragFloat3("##Scale", ref Scale, BaseSpeedSca * multiplier, axisColors);
 			ImGui.SameLine();
 			ControlButtons.ButtonChangeOperation(OPERATION.SCALE, iconScale);
 
-			ImGui.PopItemWidth();
 			IsEditing = result;
 
-			if (Ktisis.Configuration.TransformTableDisplayMultiplierInputs) {
```
Fayti1703, line 88: Any reason you dropped this `if`?

PR #62, opened 2022-12-05
Mouse input Events & deselect on void click

emyxiv:
Deselect Gizmo when click on nothing.

Known issue: the mouse click events aren't prevented when clicking on "Group Pose Settings" vanilla window.


Emyka:
Added Input methods to check if cursor is over a selectable actor or a vanilla window, "Deselect bone on click void" uses them.
Added an option to disable "Deselect bone on click void".

The PR should be ready now.

Ktisis/Interface/Input.cs
```diff
@@ -253,5 +279,44 @@ public enum Purpose {
...
+				// { "GroupPoseStampImage" }, // this seem to take almost all the screen
+			};
+		private unsafe static bool IsMouseOverWindow() {
+
+			foreach(var addonName in MouseObstructingAddonNames) {
+				var address = Services.GameGui.GetAddonByName(addonName, 1);
+				if (address == IntPtr.Zero) continue;
+
+				var addon = (FFXIVClientStructs.FFXIV.Component.GUI.AtkUnitBase*)address;
+				if (addon == null && !addon->IsVisible) continue;
```
Fayti1703, line 301: If the LHS of this `&&` is `true`, the RHS will result in a `NullReferenceException`. Did you mean this?
```suggestion
				if (addon == null || !addon->IsVisible) continue;
```

Ktisis/Interface/Input.cs
```diff
@@ -253,5 +279,44 @@ public enum Purpose {
...
+
+				var rootNode = addon->RootNode;
+				if (rootNode == null && !rootNode->IsVisible) continue;
```
Fayti1703, line 304: Same here:
```suggestion
				if (rootNode == null || !rootNode->IsVisible) continue;
```

Emyka:
Fixup several commits to clean-up git history.

Fixed the bugs mentioned by @Fayti1703.
Thanks to Ashley for the testing on this PR.

PR #63, opened 2022-12-06
Modify History directly instead of copying it

therealedited:
The current way of handling "forks" is copying the history up until the point that we want to fork. In the case of a big gpose user, it could potentially cause issues due to lots of items being copied.

### DIFF
--- a/Ktisis/History/HistoryManager.cs
+++ b/Ktisis/History/HistoryManager.cs
@@ -17,8 +17,7 @@ using Ktisis.Structs.Actor.State;
 namespace Ktisis.History {
 	public static class HistoryManager {
 		public static List<HistoryItem>? History { get; set; }
-		private static int _currentIdx = -1;
-		private static int _maxIdx = -1;
+		private static int _currentStateIndex;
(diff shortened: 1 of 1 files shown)


Ktisis/History/HistoryManager.cs
```diff
@@ -131,9 +131,8 @@ public static class HistoryManager {
 				PluginLog.Warning("Attempted to add an entry to an uninitialised history list.");
 				return;
 			}
-			History.Insert(_maxIdx, historyItem);
+			History.Insert(History.Count, historyItem);
```
Fayti1703, line 134: Use `History.Add` instead.
```suggestion
			History.Add(historyItem);
```

Fayti1703 (changes requested):
See below.

As an additional note, what exactly does `_currentIdx` describe? Based on the name, I assumed it was the index of the most-recent history item, but based on how you check against `History.Count` it looks to be the index *after* that item? Might need a rename if so.

therealedited:
See https://github.com/ktisis-tools/Ktisis/pull/64

PR #64, opened 2022-12-06
HistoryManager refactoring

therealedited:
When applied, these commits will refactor some parts of the HistoryManager, along with some minor fixes. They will also bring back the transform table saving feature.


therealedited:
Hopefully nothing important broke while rebasing with the most recent changes on main. However it would be best if it had more testing before being merged.

PR #69, opened 2022-12-08
Add opacity property for skeleton lines

Fia3D:


Ktisis/Overlay/Skeleton.cs
```diff
@@ -93,7 +94,7 @@ string Name
 					var boneColRgb = Ktisis.Configuration.GetCategoryColor(bone);
 					if (isUsing) boneColRgb.W = Math.Min(0.15f, boneColRgb.W);
 
-					var boneColor = ImGui.GetColorU32(boneColRgb);
+					var boneColor = ImGui.GetColorU32(boneColRgb.WithW(Ktisis.Configuration.SkeletonLineOpacity));
```
Fayti1703, line 97: It looks like this would overwrite the existing Alpha calculation if the Gizmo is in use above. Suggest moving the extension method call to `.GetCategoryColor` instead.

Fayti1703 (changes requested):

Ktisis/Overlay/Skeleton.cs
Fia3D: > 

I didn't put it in this method, I just multipled the alpha by the opacity

Fayti1703: I meant putting it at the end of the `.GetCategoryColor` call, not *inside* the `GetCategoryColor` method.

Fayti1703 (approved):

Fia3D:
I added a second property to handle the opacity while using the gizmo instead of hardcoded 0.15

PR #77, opened 2022-12-09
Built-in face bone position offsets 

emyxiv:
Big thanks to [Hidden by Ktisis] for the effort of making the offset for all races and genders ! ❤️ 
Original data: https://docs.google.com/document/d/1N_aUQ0jDUL1h2yzA68HvL7rwryau1h1cRY6itfPaHsE

Also added a config to disable the built-in fallback position offsets.


Maxunit:
This PR has been open for a bit over a week by now. I haven't tested it with every race yet, but heavily with my Female Highlander and Female Viera and my test results with those offsets on those two:

Highly recommended.

I will dedicate some more time to testing them with other races.

PR #78, opened 2022-12-10
Make debug solution config actually target debug

AsgardXIV:
I'm assuming this isn't intentional?
Building the debug configuration of the solution actually builds the release config.


Fayti1703:
Landed as 883d8509f8e4ef4735eba07707d3d1c30eebe8ea. Thank you.

PR #85, opened 2022-12-15
Add Bone Search

sudojunior:
- `SearchText` matches regardless of text case. (Thanks @Fayti1703)

### Changes between ab3c6a1...19e24e0

- Use of `ImGuiTreeNodeFlags` was abandoned for a new enum associated to bone relationships.
  This currently encompasses *Self* and **Child / Descendant**, but can be expanded to include:
  - Ancestor
  - _Direct_ Parent
  - _Splitting Child and Descendant_
- New enum now uses `FontAwesomeIcon` to describe its association, if any exists - but can be changed to use colours or utilize both in varying capacities.
- Icon representations:
  - Selected (current): {wand}, (descendent): 🧙🏻 
  - Queried / Searched (current): 🔍, (descendent): 🔍 (hollow)

image

Representation is expected to change to be more accessible, as is the core functionality and quality of code if this is to merge.


Ktisis/Interface/Components/BoneTree.cs
```diff
@@ -32,30 +49,49 @@ public class BoneTree {
...
-			if (Skeleton.IsBoneSelected(bone))
-				flag |= ImGuiTreeNodeFlags.Selected;
+			var criteria = HighlightCriteria.None;
 
-			var show = DrawBoneNode(bone, flag, () => OverlayWindow.SetGizmoOwner(bone.UniqueName));
-			if (show) {
+			if (isSelected) criteria ^= HighlightCriteria.Selected;
+			if (isQueried) criteria ^= HighlightCriteria.Queried;
+			if (hasChildSelected) criteria ^= HighlightCriteria.ChildSelected;
+			if (hasChildInQuery) criteria ^= HighlightCriteria.ChildQueried;
```
sudojunior, line 74: Attempts to unify this into a single method, went well initially - but hit a wall with [CS0315](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0315) when the generic was not satisfied with using the enum (the XOR operator wasn't satisfied either, which would have required each value of the enum to be type cast as an integer).
```cs
var criteria = ApplyMulti(
  HighlightCriteria.None,
  new [] {
    (HighlightCriteria.Selected, Skeleton.IsBoneSelected(bone)),
    (HighlightCriteria.Queried, BoneDoesMatch(bone)),
    (HighlightCriteria.ChildSelected, decendents.Any(BoneDoesMatch)),
    (HighlightCriteria.ChildQueried, decendents.Any(bone => Skeleton.IsBoneSelected(bone)))
  }
);

private static T ApplyMulti<T>(T initial, (T, bool)[] pairs) where T : IBitwiseOperators<T, T, T> {
  T result = initial;

  foreach (var (value, condition) in pairs) {
    if (condition) result ^= value;
  }

  return result;
}
```
The use of an array only for this, would also have been excessive for its repeated calls, overall performance and impact on garbage collection.

Fayti1703, line 74: Use OR instead of XOR here (same behavior, but with OR the intent of setting flags is clearer)

```suggestion
			if (isSelected) criteria |= HighlightCriteria.Selected;
			if (isQueried) criteria |= HighlightCriteria.Queried;
			if (hasChildSelected) criteria  |= HighlightCriteria.ChildSelected;
			if (hasChildInQuery) criteria |= HighlightCriteria.ChildQueried;
```

Ktisis/Interface/Components/BoneTree.cs
```diff
@@ -32,30 +49,49 @@ public class BoneTree {
...
+				name => name.Contains(SearchText, StringComparison.CurrentCultureIgnoreCase)
+			);
+
 		private static void DrawBoneTreeNode(Bone bone) {
 			if (bone == null) return;
 
 			var children = bone.GetChildren();
+			var decendents = bone.GetDescendants();
+
+			bool hasChildInQuery = decendents.Any(BoneDoesMatch);
```
Fayti1703, line 64: I would check for `SearchText != ""` outside of the any call here, like this:

```suggestion
			bool hasChildInQuery = SearchText != "" && decendents.Any(BoneDoesMatch);
```

Doing so prevents an unnecessary iteration over the entire array if there's no search text entered.

Ktisis/Interface/Components/BoneTree.cs
```diff
@@ -32,30 +49,49 @@ public class BoneTree {
 				}
 			}
 		}
+
+		private static bool BoneDoesMatch(Bone bone) =>
```
Fayti1703, line 53: imo this method should be renamed to `BoneMatchesSearch` or `DoesBoneMatchSearch` or similar -- with the current name it's not quite clear *what* the bone needs to match for this to return `true`.

Ktisis/Interface/Components/BoneTree.cs
```diff
@@ -32,30 +49,49 @@ public class BoneTree {
 				}
 			}
 		}
+
+		private static bool BoneDoesMatch(Bone bone) => 
+			SearchText != "" && (new [] { bone.UniqueName, bone.LocaleName }).Any(
+				name => name.Contains(SearchText, StringComparison.CurrentCultureIgnoreCase)
+			);
+
```
Fayti1703, line 57: Regarding the 'Criteria' building, you said
> The use of an array only for this, would also have been excessive for its repeated calls, overall performance and impact on garbage collection.

Isn't that also the case here? You can simply replace the array-`Any` with a boolean-OR:

```suggestion
			bone.UniqueName.Contains(SearchText, StringComparison.CurrentCultureIgnoreCase) || bone.LocaleName.Contains(SearchText, StringComparison.CurrentCultureIgnoreCase);
```

(this suggestion assumes the `SearchText != ""` check is moved outside the method)

Ktisis/Interface/Components/BoneTree.cs
```diff
@@ -32,30 +49,49 @@ public class BoneTree {
...
+			{
 				foreach (var child in children)
 					DrawBoneTreeNode(child);
 				ImGui.TreePop();
 			}
 		}
 
-		private static bool DrawBoneNode(Bone bone, ImGuiTreeNodeFlags flag, System.Action? executeIfClicked = null) {
+		private static bool DrawBoneNode(Bone bone, ImGuiTreeNodeFlags flag, HighlightCriteria criteria, System.Action? executeIfClicked = null) {
 			if (bone == null) return false;
```
Fayti1703, line 89: Is this branch ever hit? We have nullability on and `bone` is not marked as nullable...
```suggestion
```

Ktisis/Interface/Components/BoneTree.cs
```diff
@@ -7,14 +7,31 @@
 using Ktisis.Structs;
 using Ktisis.Structs.Bones;
 using Ktisis.Structs.Actor;
+using Ktisis.Util;
+using Dalamud.Interface;
+using System;
 
 namespace Ktisis.Interface.Components {
 	public class BoneTree {
+		private enum HighlightCriteria {
```
Fayti1703, line 16: I'm not sure `Criteria` is the right term to use here, though I don't know what a good alternative would be. Possibly `HighlightReason` or something to that effect?

Ktisis/Interface/Components/BoneTree.cs
```diff
@@ -32,30 +49,49 @@ public class BoneTree {
...
+			if (isQueried) criteria ^= HighlightCriteria.Queried;
+			if (hasChildSelected) criteria ^= HighlightCriteria.ChildSelected;
+			if (hasChildInQuery) criteria ^= HighlightCriteria.ChildQueried;
+
+			var flag = ImGuiTreeNodeFlags.SpanFullWidth;
+			flag |= children.Count > 0 ? ImGuiTreeNodeFlags.OpenOnArrow : ImGuiTreeNodeFlags.Leaf;
+
+			var show = DrawBoneNode(bone, flag, criteria, () => OverlayWindow.SetGizmoOwner(bone.UniqueName));
+			if (show)
+			{
```
Fayti1703, line 81: Code Style: Fold this hanging brace up.
```suggestion
			if (show) {
```

Ktisis/Interface/Components/BoneTree.cs
```diff
@@ -7,14 +7,31 @@
...
+			ChildSelected = 4,
+			ChildQueried = 16
+		}
+
 		private static Vector2 _FrameMin;
 		private static Vector2 _FrameMax;
 
+		private static string SearchText = "";
+
 		public static unsafe void Draw(Actor* actor) {
```
sudojunior, line 29: [IDE0036](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036) was raised in accordance with [`.editorconfig`](https://github.com/ktisis-tools/Ktisis/blob/238952ca54ea5fc5cfbab339e66402ab89123fff/.editorconfig#L8).
```suggestion
		public unsafe static void Draw(Actor* actor) {
```

PR #86, opened 2022-12-15
Add `GuiHelpers.FormattedString` (no visual changes) + example

sudojunior:
* Currently only supports `string` and `FontAwesomeIcon`.
* Further support towards Input handling is unlikely.

> **Note**
> I do not expect this to be merged, given the overhead needed to ensure it runs correctly on each type it would handle. However, if it does get to that stage, considerations to support other native types would be on the table.

[C# Reference](https://learn.microsoft.com/en-us/dotnet/api/system.formattablestring?view=net-6.0)

### DIFF
--- a/Ktisis/Interface/Windows/Workspace/Workspace.cs
+++ b/Ktisis/Interface/Windows/Workspace/Workspace.cs
@@ -190,11 +190,12 @@ namespace Ktisis.Interface.Windows.Workspace
 
 				if (!Categories.DrawToggleList(cfg)) {
 					ImGui.Text("No bone found.");
-					ImGui.Text("Show Skeleton (");
-					ImGui.SameLine();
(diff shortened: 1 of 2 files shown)


Fayti1703:
No offense, but I suspect this wouldn't interact well with localization either.
I'm not too familiar with the `FormattableString` API, though, so that may be incorrect.

sudojunior:
Agreed, given that we've seen this in action for SBox's Console, I thought it'd be worth a try.
With that in mind C# doesn't have the same kind of support as something like JavaScript's string templates, where string segments are provided in two arguments and also split respectively (and then recombining using `String.raw` like a zip method) - [Demo (gist.github.com)](https://gist.github.com/sudojunior/ab06cc7583726f1c521b9019c79904ad#file-year-range-ts-L16-L37).

Even if this were allowed to merge, localization would be a nightmare for the fragments it would end up having.

PR #87, opened 2022-12-17
Make ToolbarWindow toggleable via command

Fayti1703:
Baseline change as it says on the tin.

I've adjusted the GPose event listener to use `bool` while I was at it. If you want this as a separate PR (or merge just those early), tell me.

Open Issues:
- [ ] These changes are untested
- [ ] If the `toolbar` sub-command is executed outside of GPose, nothing currently happens. Unsure if we should just allow the toolbar window up regardless or display a warning to the user somehow. (I'd prefer the latter, but I don't know how to do that with Dalamud)

PR #89, opened 2022-12-18
Move Global State Init to an Attribute-based system

Fayti1703:
What it says on the tin.

Open Issues:
- [ ] These changes are not tested.
- [ ] There may be problems with ordering of the initialization.

Because I don't really trust the ordering to not become a problem, I'm marking this PR as a draft until it is field-tested.

Relations:
- Conflicts #87's "Staticify ActorStateWatcher" commit, with minor severity and arbitrary ordering (either can be rebased on the other).


Fayti1703:
@Emyka [reported](https://discord.com/channels/975894364020686878/1004359659613859910/1055851382499979345) that there currently don't appear to be any ordering issues in the existing initializations.

I would still like to add an `InitAfter` field to `GlobalStateAttribute` prior to merging this, however.

Fayti1703:
Rebased on top of current `main` and resolved conflicts.

PR #91, opened 2022-12-21
Fix: Generate category hotkeys on init

emyxiv:
Move the code responsible of generating category related purposes away from `Input.PurposesWithCategories`.
The variables `PurposesCategories`, `PurposesCategoriesHold` and `PurposesCategoriesToggle` are not populated on init.

This fixes the issue where category-based hotkeys would not work until config input tab is displayed.

Thanks to Quickdraw6 for reporting this issue.


Ktisis/Interface/Input.cs
```diff
@@ -202,25 +203,23 @@ public enum Purpose {
...
-				foreach (var category in Category.Categories) {
-					PurposesCategories.TryAdd((Purpose)i, category.Value);
-					PurposesCategoriesHold.TryAdd((Purpose)i, category.Value);
-					purposesWithCategories.Add((Purpose)i);
-					i++;
-				}
+				return Enum.GetValues<Purpose>().Concat(PurposesCategories.Keys).ToList();
+			}
+		}
+		public static void GeneratePurposesCategories() {
```
Fayti1703, line 209: This should probably be `private`.

Fayti1703 (approved):

PR #94, opened 2023-01-06
Animation, Speed Control and Status Effects

AsgardXIV:
This is an initial implementation of:
* Action/Animation control including base animation and blending
* Animation speed control /scrubbing with various different modes
* Ability to add/remove status effects from characters

The UI is pretty janky (and UI is not my strong point in general), and there are a few spots (like the inputs and gpose mode change listener) where it wasn't super clear where it best belonged.

At the very least, this is a good start on it and basically brings functional level parity (other than emote search and other UI stuff) with the Animation work I did for Anamnesis. 
The implementation is quite different to Ana in spots, using some new sigs I was working on for another project.


Fayti1703:
It looks like there are a few formatting/style problems with the changes here -- namely indentation with 4 spaces instead of tabs, hanging braces (instead of inline ones) and ordering of `using` declarations -- which also muddy the diff a little.

Could you resolve these? (Ideally you'd also set up your IDE such that it follows the existing code style -- I've been meaning to look into adjusting the `.editorconfig` to reflect the conventions here more accurately so this happens automatically, but that might take a while)

chirpxiv:
This looks amazing, thanks for the contribution! Fayti and I will be doing a quick code review so that this can get merged ASAP. It may need to go into its own branch as Ktisis is in an effective feature freeze until 0.3.

> The UI is pretty janky (and UI is not my strong point in general), and there are a few spots (like the inputs and gpose mode change listener) where it wasn't super clear where it best belonged.

This isn't any fault of your own and is mostly a result of a lack of planning ahead on my part; the v0.3/ui-rework branch aims to resolve this and should hopefully make implementation of new features a breeze. Thanks for being willing to work through my jank nonetheless! 😛

Fayti1703:
Me and *quick* code review... we'll see about that, I guess.

AsgardXIV:
I have adjusted the styling. 
It really would be ideal if you had an editor config seeing as it's so far away from the norm in C# projects.

Caraxi:
>It really would be ideal if you had an editor config seeing as it's so far away from the norm in C# projects.

there is one https://github.com/ktisis-tools/Ktisis/blob/main/.editorconfig

AsgardXIV:
> 

It's not correct though. It defines basically none of the style that is actually used, which I think is what Fayti was mentioning, it's incomplete/incorrect for now.

Fayti1703:
> It defines basically none of the style that is actually used, which I think is what Fayti was mentioning, it's incomplete/incorrect for now.

That it is -- it's quite incomplete.
It does define *some* of the style, but mostly in a way that Resharper/Rider understand, not Visual Studio.
I only even found out VS used different (and *more*) properties when I actually had to use it a week ago.

AsgardXIV:
Most of these changes made it into upstream Dalamud so I'm just gonna close this one out and reoopen.
I'll also split up Animation and Status Effects into separate PRs.

PR #99, opened 2023-01-15
CI: Setup build tests + auto release CI

BitsOfAByte:
Discussed this PR with @chirpxiv before moving forward. This PR aims to streamline the PR testing & release process for Ktisis with the help of GitHub Actions.

An example release using this CI can be found here:
https://github.com/BitsOfAByte/container-testing-ci/releases/tag/v0.7.2.3

An example build run can be found here:
https://github.com/BitsOfAByte/container-testing-ci/actions/runs/3920817168/jobs/6702655988

An example repo.json update can be found here (it uses the current repo, ignore the url change)
https://github.com/BitsOfAByte/container-testing-ci/commit/9c868eb7bb6f7c21a54f9aed9dbf387742e18451

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,62 @@
+name: Tests
+
+# If another commit is made to the same pull request, the previous workflow run is cancelled to save resources.
+concurrency:
+      group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+      cancel-in-progress: true
+
+# Only run CI against the main branch or on PRs to the main branch.
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+    types: [ready_for_review, opened, synchronize]
+  workflow_dispatch:
+
+jobs:
+  Build:
+    if: github.event.pull_request.draft == false # Ignore draft PRs
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: Ktisis/
+        shell: bash
+    strategy:
+      matrix:
+        dotnet-version: [7.0.x] # Can add multiple .NET versions here to test against (For example, testing a new version of .NET before it's released)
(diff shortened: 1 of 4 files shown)


chirpxiv:
no way I cant believe Blooym is already releasing Ktisis 0.7 for us

Ktisis/Ktisis.csproj
```diff
@@ -24,6 +24,10 @@
   <PropertyGroup>
     <DalamudLibPath>$(appdata)\XIVLauncher\addon\Hooks\dev\</DalamudLibPath>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(OS)' != 'Windows_NT' OR '$(CI)' == 'true'">
```
Fayti1703, line 27: I know I'm a bit late on this, but shouldn't this condition just be `$(CI) == 'true'`?

.github/workflows/test.yml
```diff
@@ -0,0 +1,62 @@
...
+        run: dotnet build -c Debug --no-restore --nologo
+
+      - name: Build plugin in release mode
+        run: dotnet build -c Release --no-restore --nologo
+
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: Build Artifacts
+          path: Ktisis/bin/
```
Fayti1703, line 62: Should a test build really upload artifacts?

Fayti1703: Having looked a bit more into it, this is fine. (It only drops it into the workflow results)

Ktisis/Ktisis.csproj
BitsOfAByte: It's a holdover from another project that allows me to use a VSCode devcontainer and set the same environment variable for Dalamud location, doesn't need to stay if unwanted

PR #101, opened 2023-01-15
Cleanup Workflows a little

Fayti1703:
See commit messages.

Should have no real semantic difference.


BitsOfAByte:
Noticed when reading over the testing file that my implementation of the Dalamud version matrix builds won't work due to the staging zip being stored in a different directory altogether in the dalamud-distrib repository. It might be a good idea to swap that matrix with just a regular `latest.zip`, my bad there!

PR #117, opened 2023-10-04
Auto Backup/Save poses

Cazzar:
Re creation of #114

This PR adds the ability to automatically export poses on a set interval.

This is disabled by default, but allows you set an interval and location for saving.
Some refactors were needed to make the exporting and importing features usable from multiple locations.


Fayti1703:
Closing in favor of re-opening the old PR.

PR #119, opened 2023-10-09
AutoSave - Minor Adjustments

Cazzar:
Some feedback from friends and from my own observations

Firstly I have updated the autosave to always run the thread that will poll for saving, but just never save/perform any action.

Secondly, since the default is to no longer to clear on exit, include todays date on the auto save directory.

### DIFF
--- a/Ktisis/Util/PoseAutoSave.cs
+++ b/Ktisis/Util/PoseAutoSave.cs
@@ -16,10 +16,10 @@ namespace Ktisis.Util {
 		private string SaveFolder => Ktisis.Configuration.AutoSavePath;
 
 		public void Enable() {
-            if (!Ktisis.Configuration.EnableAutoSave)
-                return;
+			if (!Ktisis.Configuration.EnableAutoSave)
+				return;
 
-            if (!Directory.Exists(SaveFolder))
+			if (!Directory.Exists(SaveFolder))
 				Directory.CreateDirectory(SaveFolder);
 
 			_timer = new Timer(TimeSpan.FromSeconds(Ktisis.Configuration.AutoSaveInterval));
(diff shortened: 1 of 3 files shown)


Ktisis/Util/PoseAutoSave.cs
chirpxiv, line 1: I believe this could be better addressed by adding a check when enabling autosaves, ie.
```cs
if (enableAutoSave && PoseHooks.PosingEnabled)
	PoseHooks.AutoSave.Enable();
```

Also worth mentioning that a similar issue exists when updating the autosave interval; if active while being changed, then the timer will wait for the previously configured interval to complete. It may be worth adding a method to `PoseAutoSave` to update the timer for configuration changes?

Cazzar: Admittedly I was tossing up between the two approaches - the reason for that approach was more thinking to make it easier than having to check everywhere.

With the timer, it's a bit of a lower case issue for it there, but I agree it is there, I will have to look at the side effects of changing the interval

Cazzar: From [MSDN](https://learn.microsoft.com/en-us/dotnet/api/system.timers.timer.interval?view=net-7.0):
> If the interval is set after the [Timer](https://learn.microsoft.com/en-us/dotnet/api/system.timers.timer?view=net-7.0) has started, the count is reset. For example, if you set the interval to 5 seconds and then set the [Enabled](https://learn.microsoft.com/en-us/dotnet/api/system.timers.timer.enabled?view=net-7.0) property to true, the count starts at the time [Enabled](https://learn.microsoft.com/en-us/dotnet/api/system.timers.timer.enabled?view=net-7.0) is set. If you reset the interval to 10 seconds when count is 3 seconds, the [Elapsed](https://learn.microsoft.com/en-us/dotnet/api/system.timers.timer.elapsed?view=net-7.0#system-timers-timer-elapsed) event is raised for the first time 13 seconds after [Enabled](https://learn.microsoft.com/en-us/dotnet/api/system.timers.timer.enabled?view=net-7.0) was set to true.

Given that side effect Chirp, do you think that's reasonable for us to update the Interval when the value gets updated? It could cause some other complications.

Currently it's handled in the here: https://github.com/ktisis-tools/Ktisis/blob/0969bb65f17526b4be5b13e6afd519c38fb6c50c/Ktisis/Util/PoseAutoSave.cs#L88-L89

Fayti1703: > the reason for that approach was more thinking to make it easier than having to check everywhere.

It might be worth consolidating the runtime update logic in a method, then. That way you'd only need to call that single method when the configuration is changed.

chirpxiv: > Given that side effect Chirp, do you think that's reasonable for us to update the Interval when the value gets updated? 

This seems like the expected behaviour for setting a new interval - it resets the timer and starts counting towards the newly set one, which should also make the check at the end of `Save` redundant 👍

Cazzar: Thanks for all that feedback - Updated to add a central `PoseHooks.AutoSave.UpdateSettings()` which handles the following cases:

1. Enabling Auto Save with posing enabled.
2. Disabling Auto Save with posing enabled
3. Updating timer

Fayti1703:
f92d4c26b83a6b5f5b1664fa31dd01d9afe2ca6d seems to introduce a lot of indentation issues. Can you check your IDE settings and correct the indentation?

Ktisis/Util/PoseAutoSave.cs
chirpxiv: Looking over it now, thank you!

Cazzar:
> [f92d4c2](https://github.com/ktisis-tools/Ktisis/commit/f92d4c26b83a6b5f5b1664fa31dd01d9afe2ca6d) seems to introduce a lot of indentation issues. Can you check your IDE settings and correct the indentation?

Yeah.. sorry, my IDE seems to want to preference 4 spaces, let me fix that up, I keep forgetting to check for this.

Seems the editorconfig doesn't define it there. 
image

chirpxiv:
Looks good, ready to merge this if you're happy with it!

Fayti1703:
Landed as [`dc288825`..`4c5404cd`](https://github.com/ktisis-tools/Ktisis/compare/0969bb65f17526b4be5b13e6afd519c38fb6c50c...dc288825867a29eccab2129c690c0f7a4d8bdafd). Thank you for your contribution!

PR #122, opened 2023-11-15
Add ability to load .cmp files

Spiderbuttons:
This adds the ability to load older poses in .cmp format by converting them to .pose format after the pose file string is loaded on import but before JSON deserialization. Bone names are not converted during this step as the existing legacy bone name conversion will handle it.

### DIFF
--- a/Ktisis/Helpers/PoseHelpers.cs
+++ b/Ktisis/Helpers/PoseHelpers.cs
@@ -54,7 +54,7 @@ namespace Ktisis.Helpers {
 
 		public unsafe static void ImportPose(Actor* actor, List<string> path, PoseMode modes) {
 			var content = File.ReadAllText(path[0]);
-			if (path[0].Contains(".cmp")) content = LegacyPoseHelpers.ConvertLegacyPose(content);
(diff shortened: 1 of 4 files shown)


Ktisis/Helpers/PoseHelpers.cs
```diff
@@ -53,6 +54,7 @@ internal class PoseHelpers {
 
 		public unsafe static void ImportPose(Actor* actor, List<string> path, PoseMode modes) {
 			var content = File.ReadAllText(path[0]);
+			if (path[0].Contains(".cmp")) content = LegacyPoseHelpers.ConvertLegacyPose(content);
```
chirpxiv, line 57: I recommend using the [Path.GetExtension](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getextension) method for this

chirpxiv (changes requested):
Looks good, thank you for taking the time to implement this!

Only two minor nitpicks here - in addition to what else was mentioned, could you format `LegacyPoseHelpers` to use K&R / 1TBS style indentation? Meaning new lines are not made for brackets, including the opening braces of functions.

Other than that, this looks good to merge once I've had the chance to test it. Thanks again!

chirpxiv (approved):

PR #151, opened 2024-08-06
Add Skelomae bone categorisation

Cazzar:
This adds the bones that is added in Skelomae to categories. 
Bones are categorised based on both the specific type of creature that they are looking to augment, as well as the location as to where they are

<details>
<summary>An example of the changes ingame</summary>

image
</details>


Cazzar:
I might work on trying to make the migration a little better - this seems to have reset my category colour settings.

PR #159, opened 2024-09-01
add optional localization tickbox to config.Categories

sleepybnuuy:
this may be a small use case so review at your discretion, but in the course of modding users will often need to check the weights/visibility of particular bones in the armature by referencing their formal SE/JP name instead of a localized one. this PR adds a checkbox (defaulted to true) to always show localized names where available when in use, which preserves the current behavior. when ticked off, untranslated names (ala j_kami, j_kao, iv_sippo_a) will be shown for bones in the node list and overlay instead. if there's ever an Experimental/goof-proof settings tab, this might belong in there instead

image

### DIFF
--- a/Ktisis/Interface/Windows/ConfigWindow.cs
+++ b/Ktisis/Interface/Windows/ConfigWindow.cs
@@ -94,6 +94,8 @@ public class ConfigWindow : KtisisWindow {
 		this.DrawHint("config.categories.hint_nsfw");
 		refresh |= ImGui.Checkbox(this.Locale.Translate("config.categories.show_all_viera_ears"), ref this.Config.Categories.ShowAllVieraEars);
 		this.DrawHint("config.categories.hint_viera_ears");
+		refresh |= ImGui.Checkbox(this.Locale.Translate("config.categories.show_friendly_bone_names"), ref this.Config.Categories.ShowFriendlyBoneNames);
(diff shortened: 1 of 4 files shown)


Ktisis/Interface/Windows/ConfigWindow.cs
```diff
@@ -94,6 +94,8 @@ private void DrawCategoriesTab() {
 		this.DrawHint("config.categories.hint_nsfw");
 		refresh |= ImGui.Checkbox(this.Locale.Translate("config.categories.show_all_viera_ears"), ref this.Config.Categories.ShowAllVieraEars);
 		this.DrawHint("config.categories.hint_viera_ears");
+		refresh |= ImGui.Checkbox(this.Locale.Translate("config.categories.show_friendly_bone_names"), ref this.Config.Categories.ShowFriendlyBoneNames);
+		this.DrawHint("config.categories.hint_friendly_bones");
```
Fayti1703, line 98: A refresh necessary here? It doesn't look like the bone names are cached.

```suggestion
		ImGui.Checkbox(this.Locale.Translate("config.categories.show_friendly_bone_names"), ref this.Config.Categories.ShowFriendlyBoneNames);
		this.DrawHint("config.categories.hint_friendly_bones");
```

PR #189, opened 2025-09-03
reinstate gaze controls pt1

sleepybnuuy:
largely attempts to reimplement most/all functionality from the main release. probably has some very hacky solutions that should be fixed

### Changes
- ActorPropertyList.cs
    - logic for drawing actor gazes on the object editor's dropdowns. supports both Linked (All gaze parts simultaneously) and Unlinked (Head, Eyes, Torso parts separately) control.
    - does not implement gizmo selection/visible transforms. camera tracking works fine, so i think this could be released without too much pain, but the gizmo drawing logic is a lot more restricted in v0.3 compared to main. we'd probably want to generate a dummy TargetGaze object in the scene to point to, then use the overlay's native gizmo to drag it around as needed. alt., could implement similar to IK bones, but that seems like a mess
- ActorEntity.cs
    - added a nullable ActorGaze here so we can store&reference edited gazes on each ActorEntity in our scene without ActorPropertyList maintaining a big dictionary of them (EditGaze does this on the main branch and i dislike it)
- ActorModule.cs
    - re-adds the ActorLookAtDelegate and ControlGazeDetour; i'm still on the fence about whether or not we even need ControlGazeDetour, or if we could consistently call ActorLookAtDelegate with our custom gazes whenever necessary.
- ActorGaze.cs (new)
    - essentially ported 1:1 from the main branch - this has heavy overlap with ClientStructs' [LookAtContainer](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Client/Game/Character/LookAtContainer.cs#L12) and related entities, so it might be worthwhile to just extend those or ape them where we can.
- CharacterEx.cs
    - added the character offset to point at an ActorGaze for any characters in scene

### Todo
- convert types to ClientStructs equivalents?
- probably some memory cleanup, theres a lot of vars just floating around here
- should gaze generation/application logic be further separated from DrawGaze & DrawGazeTab?
- reinstate gaze gizmo
- ~~fix hook warning in ActorModule~~ side effect of calling CheckValid, going to ignore
- ~~logic here assumes vanilla skp files for affected players; what happens when you try to change torso gaze on a modded skp that has no torso movement?~~ gaze hooking follows the constraints of a modded skp file; nothing we can really do about this atm

GIF 9-2-2025 3-45-54 PM

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,9 +1,4 @@
 
-[*.cs]
-indent_style = tab
-indent_size = tab
-tab_width = 4
-
 [*]
 
 # Microsoft .NET properties
@@ -71,3 +66,14 @@ resharper_wrap_before_primary_constructor_declaration_rpar = true
 resharper_wrap_chained_binary_patterns = chop_if_long
 resharper_wrap_list_pattern = chop_always
 resharper_wrap_object_and_collection_initializer_style = chop_always
+
+
+[*.cs]
+indent_style = tab
+indent_size = tab
+tab_width = 4
+
+[*.json]
+indent_style = tab
+indent_size = tab
+tab_width = 4
--- a/Ktisis/Data/Config/Bones/BoneCategory.cs
+++ b/Ktisis/Data/Config/Bones/BoneCategory.cs
@@ -19,4 +19,6 @@ public class BoneCategory(string name) {
 	
 	public TwoJointsGroupParams? TwoJointsGroup;
 	public CcdGroupParams? CcdGroup;
+
+	public List<string> Presets = new();
 }
--- a/Ktisis/Data/Config/ConfigManager.cs
+++ b/Ktisis/Data/Config/ConfigManager.cs
@@ -1,7 +1,9 @@
 using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.IO;
 using System.Diagnostics;
-
+using System.Linq;
 using Dalamud.Plugin;
 
 using Newtonsoft.Json;
@@ -43,6 +45,10 @@ public class ConfigManager : IDisposable {
 			if (cfg is { Version: < 10 }) {
 				cfg.Version = 10;
 				this.MigrateSchema(cfg);
+			}
+			if (cfg is { Version: < 11 }) {
+				cfg.Version = 11;
+				this.GenerateDefaultPresets(cfg);
 			}
 		} catch (Exception err) {
 			Ktisis.Log.Error($"Failed to load configuration:\n{err}");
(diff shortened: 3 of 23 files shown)


Ktisis/Interface/Editor/Properties/ActorPropertyList.cs
```diff
@@ -71,7 +89,111 @@ private void DrawActorTab(ActorEntity actor) {
...
+
+		// work from existing gaze on ActorEntity or make a new one if its our first touch w them
+		var gaze = actor.Gaze != null ? (ActorGaze)actor.Gaze : new ActorGaze();
+		// if human actor, enable link/unlinked controls
+		bool isHuman = actor.GetHuman() != null;
+
+		var spacing = ImGui.GetStyle().ItemInnerSpacing.X;
+		var result = false;
+
+		using (var _disable = ImRaii.Disabled(this._ctx.Posing.IsEnabled)) {
```
Cazzar, line 104: if you're not using the `_disable` property, you can simply do a discard assignment instead
```suggestion
		using (var _ = ImRaii.Disabled(this._ctx.Posing.IsEnabled)) {
```

but given this is a traditional using statement, no assignment is needed.
```suggestion
		using (ImRaii.Disabled(this._ctx.Posing.IsEnabled)) {
```

Ktisis/Interface/Editor/Properties/ActorPropertyList.cs
```diff
@@ -71,7 +89,111 @@ private void DrawActorTab(ActorEntity actor) {
...
+		}
+		ImGui.SameLine(0, spacing);
+
+		// TODO: gizmo, needs work to generate a new one w/o using ObjectWindow/OverlayWindow
+		// 	or, possible to create a dummy scene object (GazeTarget?) to hijack overlay gizmo?
+		using (var _disable = ImRaii.Disabled()) {
+			using (ImRaii.PushColor(ImGuiCol.Button, ImGui.GetColorU32(ImGuiCol.ButtonActive), isTracking)) {
+				if (Buttons.IconButtonTooltip(FontAwesomeIcon.LocationArrow, "Gizmo Tracking [TODO]", Vector2.Zero)) {}
+			}
+		}
```
Cazzar, line 193:
```suggestion
		using (ImRaii.Disabled()) {
			using var _ = ImRaii.PushColor(ImGuiCol.Button, ImGui.GetColorU32(ImGuiCol.ButtonActive), isTracking)
			if (Buttons.IconButtonTooltip(FontAwesomeIcon.LocationArrow, "Gizmo Tracking [TODO]", Vector2.Zero)) {}
		}
```

to reduce nesting for now.

Cazzar (changes requested):

PR #192, opened 2025-09-04
object editor enhancements

sleepybnuuy:
### changes

- allows toggling windows open and shut with their workspace buttons
  - when this flag is OFF, we now ImGui.SetWindowFocus the clicked window button to bring it to the front
- adds selected entity's name to header name of windows where contextually appropriate
- stub that adds pose import and chara import windows to object editor dropdowns
- prevents dragging of non-draggable scene list entries
- removes the requirement to toggle EntityPose entries open to reach bone categories
- allows selected EntityPose to render object editor context tabs for actor & pose
- adds a clear/reset button for the import npc selector, as well as better contextual disabling of import modes
- adds Reference Pose, Stash Pose, and Apply Pose buttons to the pose propertylist
  - Reference Pose is simply ported from the pose dropdown option
  - Stash Pose saves the current entitypose as a pseudo-memento on PosingManager; currently its just doing all bones and position+rotation
  - Apply Pose takes the StashedPose and applies the memento onto the currently-targeted EntityPose; this allows you to clipboard poses to pass between actors or temp-save if you're trying out different things.
  - Apply Pose has a tooltip displaying the time of the last stash and the actor the pose came from
- adds a tooltip to DrawTargetLabel listing, when hovered, any addtl targeted entities that weren't displayed in the main text
- adds a Hide/Unhide button (leveraging Character->Opacity) onto ActorEntities in the SceneTree (as well as their right-click menus)
- adds a Sibling Link selector button to the object editor gizmo pane; when a valid bone is selected (e.g. arm_L), clicking it will multiselect its sibling bone (e.g. arm_R)
- adds a config toggle to control closing object editor when nothing is selected (child setting of ToggleEditorOnSelect)

this can be broken up if we want it into chunks, but response to everything in here has so far been positive.


Ktisis/Editor/Posing/PosingManager.cs
```diff
@@ -205,6 +209,45 @@ public Task ApplyPoseFile(
...
+				Transforms = transforms,
+				Bones = null,
+				Initial = container, // todo: does initial state matter for the stash save?
+				Final = container
+			};
+			this.StashedAt = DateTime.Now;
+			this.StashedFrom = pose.Parent.Name;
+		});
+	}
+	public Task ApplyStashedPose(EntityPose pose) {
```
Cazzar, line 232: Formatting typo?
```suggestion

	public Task ApplyStashedPose(EntityPose pose) {
```

Ktisis/Interface/Components/Chara/CharaImportUI.cs
```diff
@@ -35,7 +37,9 @@ FileSelect<CharaFile> select
 	// Initialization
 
 	public void Initialize() {
```
Cazzar, line 39: is there any performance downsides to refactoring this to remove the two-stage init?

I personally find this quite an antipattern, and it's possible that one-time init might be enough for this, instead of going by the each UI load

Ktisis/Interface/Components/Chara/CharaImportUI.cs
```diff
@@ -129,21 +139,23 @@ private void DrawMethodRadio(string label, LoadMethod method) {
...
-		
-		ImGui.Spacing();
-		
-		ImGui.Text("Equipment");
-		this.DrawModeSwitch("Gear", SaveModes.EquipmentGear);
-		ImGui.SameLine();
-		this.DrawModeSwitch("Accessories", SaveModes.EquipmentAccessories);
-		ImGui.SameLine();
-		this.DrawModeSwitch("Weapons", SaveModes.EquipmentWeapons);
+		using (ImRaii.Disabled(this.DisableModes)) {
```
Cazzar, line 142: to reduce nesting you can use block level using, ex
```suggestion
		using var _ = ImRaii.Disabled(this.DisableModes);
```

Ktisis/Interface/Components/Transforms/TransformTable.cs
```diff
@@ -115,7 +119,7 @@ public bool DrawPosition(ref Vector3 position, TransformTableFlags flags = Trans
 
 	private bool DrawPosition(ref Vector3 pos, bool op) {
 		var result = this.DrawLinear("##TransformTable_Pos", ref pos);
-		if (op) this.DrawOperation(PositionOp, FontAwesomeIcon.LocationArrow, "transform.position");
+		if (op) this.DrawOperation(PositionOp, FontAwesomeIcon.LocationArrow, this._locale.Translate("transform.position"));
```
Cazzar, line 122: instead of passing in `this._locale.Translate(...)` all the time - is it worth refactoring this closer to the `DrawOperation`?

Ktisis/Interface/Components/Workspace/SceneDragDropHandler.cs
```diff
@@ -37,6 +37,7 @@ public void Handle(SceneEntity entity) {
 	}
 
 	private void HandleSource(SceneEntity entity) {
+		if (!(entity is IAttachable)) return;
```
Cazzar, line 40:
```suggestion
		if (entity is not IAttachable) return;
```

Ktisis/Interface/Components/Workspace/SceneTree.cs
```diff
@@ -130,24 +133,26 @@ private void DrawNode(SceneEntity node) {
...
+
         // Caret
 
+		var expand = false;
 		var style = ImGui.GetStyle();
 		ImGui.SameLine();
 		ImGui.SetCursorPosX(pos.X - style.ItemSpacing.X);
-		var expand = this.DrawNodeCaret(display.Color, flag);
-		
+		if(!(item is EntityPose))
```
Cazzar, line 148:
```suggestion
		if(item is not EntityPose)
```

Ktisis/Interface/Editor/Properties/PosePropertyList.cs
```diff
@@ -45,22 +49,54 @@ public override void Invoke(IPropertyListBuilder builder, SceneEntity entity) {
...
+		if (actor is not ActorEntity) return;
 		ImGui.Spacing();
 		
-		if (ImGui.Button("Import"))
-			this._ctx.Interface.OpenPoseImport(actor);
-		ImGui.SameLine(0, spacing);
-		if (ImGui.Button("Export"))
+		// if (ImGui.Button("Import"))
+		// 	this._ctx.Interface.OpenPoseImport(actor);
+		// ImGui.SameLine(0, spacing);
```
Cazzar, line 65:
```suggestion
```

PR #201, opened 2025-09-14
allow enabling presets by default

sleepybnuuy:
### changes
- adds a GPose Default checkbox to preset list entries, which controls their addition to a DefaultPresets list on preset config
- when drawing actor entities, attempt once (when their bones exist to reference) to set active any default presets
  - we do this in Draw instead of init since gameobject data may need some time to propagate; see when entering gpose for the first time, or when creating new actors

image

### DIFF
--- a/Ktisis/Data/Config/Sections/PresetConfig.cs
+++ b/Ktisis/Data/Config/Sections/PresetConfig.cs
@@ -9,9 +9,9 @@ public class PresetConfig {
 	internal static PresetRemoved? PresetRemovedEvent;
 
 	public SortedDictionary<string, ImmutableHashSet<string>> Presets = new ();
-	public List<string> DefaultPresets = new List<string>();
+	public HashSet<string> DefaultPresets = new ();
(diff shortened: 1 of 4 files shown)


Ktisis/Data/Config/Sections/PresetConfig.cs
```diff
@@ -1,11 +1,17 @@
...
-public class PresetConfig {
+public class PresetConfig {
 	internal delegate void PresetRemoved(string presetName);
 	internal static PresetRemoved? PresetRemovedEvent;
 
 	public SortedDictionary<string, ImmutableHashSet<string>> Presets = new ();
+	public List<string> DefaultPresets = new List<string>();
+
+	public bool PresetIsDefault(string name) {
+		return this.DefaultPresets.FirstOrDefault(x => x == name) != null;
```
Cazzar, line 15:
```suggestion
		return this.DefaultPresets.Contains(name);
```

PR #205, opened 2025-09-15
Allow workspace to initialize when Gpose is already active

sudojunior:
Adds a check to show `WorkspaceWindow` when Ktisis starts in GPose.

https://github.com/user-attachments/assets/3ce03751-c4f9-4ab1-ae29-b563f5994cba

It should be noted that compiling from source while in GPose has presented some problems for this, but toggling the plugin state (as shown in the demo) has not interfered with this _yet_.

- When reviewing the comparisons in functionality from v0.2, it would not be possible to show the workspace window when GPose is not active.
- The `/ktisis` chat command also fails to recognize when the context is available because it is created with the editor, which only happens if `OpenOnEnterGPose` is true.

### DIFF
--- a/Ktisis/Editor/Context/ContextManager.cs
+++ b/Ktisis/Editor/Context/ContextManager.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Diagnostics;
 
+using Dalamud.Plugin.Services;
+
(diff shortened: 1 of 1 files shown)


sleepybnuuy:
working today on v0.3/main

PR #208, opened 2025-09-16
custom pose view image support

sleepybnuuy:
### Changes
- adds PoseViewConfig and a Pose View tab to the config editor
- adds filepath inputs via file dialog picker in the Pose View config tab
- allows PoseViewRenderer to either fetch images from /Data/ or instead from PoseViewConfig values if their filepaths have been defined.
- readds the OpenBrowser functionality from 0.2's Common util helper to provide a link button to a store of template images for users to edit with their own screenshots

### Notes
- branch was originally based off v0.3/poseview-lumbar, so history includes reverts for those commits. may want to make sure all works fine in terms of merging as we add each
- will need a followup PR to point the placeholder in localization onto the actual repo templates folder
- better lips/face template? these were free-handed since magic wand was NOT helping on those images

image


Ktisis/Localization/Data/en_US.json
```diff
@@ -312,6 +312,23 @@
...
+			"ears": "Ears (Viera)",
+			"face": "Face",
+			"hands": "Hands",
+			"lips": "Lips",
+			"mouth": "Mouth",
+			"tail": "Tail",
+			"linkout": {
+				"description": "Download template images here:",
+				"button": "GitHub",
+				"link": "https://github.com/sleepybnuuy/ktisis-testing/tree/v0.3/custom-paperdolls/Ktisis/Data/Images/Templates"
```
Cazzar, line 330: Should we change this link?

PR #222, opened 2025-10-03
Refactor manual ImGui push/pop to ImRaii

Cazzar:
Migrate from using manual push/pop to ImRaii to defend against any errors mid execution, as these should auto clean up


sleepybnuuy (approved):
everything loads up good at a glance

PR #233, opened 2025-11-02
Draft: Add IPC support for bone matrix manipulation

Cazzar:
This adds the following IPCs:

- `Ktisis.SetMatrix`  input: `uint actorIndex, string boneName, Matrix4x4 matrix` returning `Task<bool>`
  - This will take a given input matrix and apply it to the given bone
  - The usage of the Matrix4x4 is to provide all transforms in one Object, you can derive position (Vec3), rotation (quaternion) and scale (vec3) from this.
- `Ktisis.GetMatrix` input: `uint actorindex, string bonename` returning `Task<Matrix4x4?>`
  - This allows retrieving current matrix.
- `Ktisis.SelectedBones` no input, returning `Dictonary<int, HashSet<string>>`
  - This provides a list of ALL selected bones across each actor.

Current bug to be fixed: when calling `Ktisis.SetMatrix` it will successfully set the bone's matrix, but it will not propogate down any other bones if required, which kinda kills the usage of it.

### DIFF
--- a/Ktisis/Editor/Transforms/TransformHandler.cs
+++ b/Ktisis/Editor/Transforms/TransformHandler.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Linq;
 using System.Numerics;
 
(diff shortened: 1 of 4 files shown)


sleepybnuuy (approved):
lgtm with matrix stuff still as a wip

Cazzar:
There's more work that needs to be done with actually implementing the matrix based functions, but, for now, that will be added as no-ops.

They will be considered unsupported and subject to change until they actually work.

The current implementation does propagate, but it doesn't seem to decompose correctly.

Merging this  in as-is for now.

PR #234, opened 2025-11-03
add pose flipping

priprii:


Ktisis/Editor/Posing/Data/EntityPoseConverter.cs
```diff
@@ -193,4 +196,77 @@ private IEnumerable<PartialBoneInfo> GetBoneSelectionFrom(IEnumerable<SkeletonNo
...
+
+		var skeleton = target.GetSkeleton();
+		if(skeleton == null || skeleton->PartialSkeletons == null)
+			return null;
+
+		var result = new PoseContainer();
+		foreach(var bone in bones) {
+			if(bone.Name.StartsWith("j_f_") || bone.Name.StartsWith("n_f_")) continue; // Skip face
+			result[bone.Name] = new Transform();
+		}
```
sleepybnuuy: you can exclude bones from the face skeleton (and hair skeleton) by skipping PartialSkeletons at index 1/2 instead of filtering by bone name strings, we do this in PoseContainer:

https://github.com/ktisis-tools/Ktisis/blob/2ed411560b5c3d6f4d8aedbc71de8edf8a2e4e80/Ktisis/Editor/Posing/Data/PoseContainer.cs#L70-L75

Ktisis/Editor/Posing/PosingManager.cs
```diff
@@ -273,7 +273,25 @@ public Task ApplyStashedPose(EntityPose pose) {
 			this.StashedPose = null;
 		});
 	}
-	
+
+	public Task ApplyFlipPose(EntityPose pose) {
```
sleepybnuuy: separate from the below suggestion, Redoing the memento can spin the actor 180 degrees. not sure where this is coming from; root rotation/transforming?

Ktisis/Editor/Posing/Data/EntityPoseConverter.cs
```diff
@@ -193,4 +196,77 @@ private IEnumerable<PartialBoneInfo> GetBoneSelectionFrom(IEnumerable<SkeletonNo
...
 	}
+
+	public unsafe PoseContainer? FlipPose() {
+		var bones = this.GetBones().ToList();
+		if(bones.All(b => b.PartialIndex == 0))
+			return null;
+
+		var skeleton = target.GetSkeleton();
+		if(skeleton == null || skeleton->PartialSkeletons == null)
+			return null;
```
sleepybnuuy: this is already checked in GetBones, so you can check if bones is an empty list and break if so

Ktisis/Editor/Posing/Data/EntityPoseConverter.cs
```diff
@@ -193,4 +196,77 @@ private IEnumerable<PartialBoneInfo> GetBoneSelectionFrom(IEnumerable<SkeletonNo
...
+
+		var result = new PoseContainer();
```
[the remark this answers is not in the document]
sleepybnuuy: this would also let you drop the somewhat-hacky `result[boneName]` indexing in favor of letting PoseContainer.Save handle all of that. i'd recommend it

Ktisis/Editor/Posing/Data/EntityPoseConverter.cs
```diff
@@ -193,4 +196,77 @@ private IEnumerable<PartialBoneInfo> GetBoneSelectionFrom(IEnumerable<SkeletonNo
...
+
+		HavokPosing.SetModelTransform(rootPose, rootBone.BoneIndex, rootTarget);
+		HavokPosing.Propagate(skeleton, rootBone.PartialIndex, rootBone.BoneIndex, rootTarget, rootCurrent);
+
+		return result;
+	}
+
+	private float GetYaw(Quaternion q) {
+		return MathF.Atan2(2f * (q.W * q.Y + q.X * q.Z), 1f - 2f * (q.Y * q.Y + q.Z * q.Z));
+	}
```
sleepybnuuy: doesnt quite belong here; can move this to a math helper like `HkaEulerAngles.cs`

sleepybnuuy (changes requested):
thank you for the PR!! currently, flipping does not seem to play nice with viera ear bones or hair skeletons (see below) so i've left a few suggestions that may help. i'd also like to see FlipPose made into a `void` return, with `ApplyFlipPose` handling the posecontainer creation with converter.Save after it runs. undo/redo memento behavior should also be touched up

image

Ktisis/Interface/Windows/ObjectWindow.cs
```diff
@@ -166,22 +167,34 @@ private void DrawToggles(ITransformTarget? target) {
...
+				using var _ = ImRaii.Disabled(!siblingAvailable || selectionCount != 1); // disable if current bone has no sibling or if multiple selections
+				if (Buttons.IconButtonTooltip(FontAwesomeIcon.PeopleArrows, siblingHint, iconBtnSize))
+					this._ctx.Selection.Select(siblingNode, SelectMode.Multiple); // if a sibling exists, select it assuming SelectMode.Multiple
+
+				ImGui.SameLine(0, spacing);
+			}
+
+			if (pose != null) {
+				if (Buttons.IconButtonTooltip(FontAwesomeIcon.ArrowRightArrowLeft, this._ctx.Locale.Translate("transform_edit.flip_pose"), iconBtnSize))
+					this._ctx.Posing.ApplyFlipPose(pose);
```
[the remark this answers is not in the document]
priprii: I think the preference of this would vary depending on usecase.

In my own version I additionally made the pose flipping functionality a keybind because I flip frequently to check how the pose looks mirrored, it helps with seeing whether anything looks weird, it's something a lot of artists like to do.

We could move it down in the pose menu for those who are only using it occasionally, while also adding a keybinding for those who require more frequent use of it?

PR #246, opened 2025-11-29
Fix Matrix Stuff and add Batch get/set

NanaKhide:
I think that works. U probably want to triple check, change debug window cause i quickly threw it together and have no solution for testing batch set

### DIFF
--- a/Ktisis/Interop/Ipc/IpcProvider.cs
+++ b/Ktisis/Interop/Ipc/IpcProvider.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


Ktisis/Interop/Ipc/IpcProvider.cs
```diff
@@ -7,187 +7,299 @@
...
+		IpcIsPosing.RegisterFunc(IsActive);
+		IpcLoadPose.RegisterFunc(LoadPose);
+		IpcLoadPoseExtended.RegisterFunc(LoadPose);
+		IpcSavePose.RegisterFunc(SavePose);
+		IpcGetMatrix.RegisterFunc(GetMatrix);
+		IpcSetMatrix.RegisterFunc(SetMatrix);
+		IpcSelectedBones.RegisterFunc(SelectedBones);
+		IpcBatchGetMatrix.RegisterFunc(BatchGetMatrix);
+		IpcBatchSetMatrix.RegisterFunc(BatchSetMatrix);
+		IpcGetAllMatrices.RegisterFunc(GetAllMatrices);
```
anya-hichu: Should call `UnregisterFunc` on Dispose/Unload so the call gates are removed

Ktisis/Common/Utility/Transform.cs
```diff
@@ -73,7 +77,41 @@ out var pos
 		this.Rotation = rot;
 		this.Scale = scl;
 	}
-	
+
+	public void DecomposeMatrixPrecise(Matrix4x4 mx, Transform initial) {
+		const float posEps = 1e-4f;
+		const float rotAngleEps = 1e-4f;
+		const float scaRelEps = 1e-4f;
```
anya-hichu: should probably just be called `eps` since always the same value

anya-hichu (commented):
boop, here are some dum comments because can't test or validate math/pose logic (LGTM tho)

Ktisis/Interop/Ipc/IpcProvider.cs
Cazzar: Looking at the diff, a lot of this isn't in Leyla's change - this is something that's a problem in the original version right now.

PR #256, opened 2025-12-14
Add flip pose back to keybinds

priprii:


sleepybnuuy:
lgtm, forgot to add the pose_ prefix when we put in the new input

PR #269, opened 2026-01-14
Pose import changes

priprii:
- Selective import with descendants (respects modes)
- Fix face expression import
- Warn for incompatible face import

Selective import rotation is parent relative, maybe implement toggle option for behaviour of this

### DIFF
--- a/Ktisis/Interface/Windows/Import/PoseImportDialog.cs
+++ b/Ktisis/Interface/Windows/Import/PoseImportDialog.cs
@@ -5,6 +5,8 @@ using Dalamud.Interface;
 using Dalamud.Interface.Colors;
 using Dalamud.Interface.Utility.Raii;
 
+using GLib.Widgets;
+
 using Ktisis.Data.Config;
 using Ktisis.Data.Files;
 using Ktisis.Editor.Context;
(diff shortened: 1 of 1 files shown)


Ktisis/Interface/Windows/Import/PoseImportDialog.cs
```diff
@@ -111,7 +113,17 @@ private void DrawApplyModes(bool isSelectBones) {
 				file.ImportPoseSelectedBones ^= true;
 		}
 
-		if (!isSelectiveImport) {
+		if (isSelectiveImport) {
+			ImGui.Indent();
```
sleepybnuuy: ImRaii has a PushIndent we could implement here as a `using` var instead
https://dalamud.dev/api/Dalamud.Interface.Utility.Raii/Classes/ImRaii.Indent

Ktisis/Editor/Posing/Data/PoseContainer.cs
```diff
@@ -56,6 +57,7 @@ public unsafe void Store(
 
 				var name = skeleton->Bones[i].Name.String;
 				if (name.IsNullOrEmpty()) continue;
+				if(filter != null && (!filter.ContainsKey(name) || p == 4)) continue;
```
sleepybnuuy: ditto helmet/body index concern above - what does p==4 represent here?

sleepybnuuy: reference pose PR #262 adds both the ability to reset specific skeletons on demand, and to automatically trigger a refpose on a skeleton slot if we detect its changed from gear/glam being applied. it should catch a lot of these cases or at least enable users to more easily fix them

PR #289, opened 2026-04-23
Tech toolbar

Glorou:
fixed some messy interactions with left clicking


sleepybnuuy (approved):
works good barring the persistent dalamud-stg issue, note in eventual patchnotes that this changes the actor editor hotkey to middleclick instead of doubleclick

PR #304, opened 2026-06-02
Migration assistant

Glorou:
This adds utilities to help migrate users from v2 to v3, or from the v3 beta to v3 stable.

### DIFF
--- a/Ktisis/Common/Extensions/GameObjectEx.cs
+++ b/Ktisis/Common/Extensions/GameObjectEx.cs
@@ -21,6 +21,8 @@ using Ktisis.Core.Types;
 using Ktisis.Core.Attributes;
 using Ktisis.Actions.Types;
 
+using ObjectType = FFXIVClientStructs.FFXIV.Client.Graphics.Scene.ObjectType;
+
 namespace Ktisis.Common.Extensions;
 
 public static class GameObjectEx {
--- a/Ktisis/Core/PluginContext.cs
+++ b/Ktisis/Core/PluginContext.cs
@@ -62,7 +62,7 @@ public class PluginContext : IPluginContext {
 		if (this.Config.GetConfigFileExists()) {
 			this.Config.Load();
 			if (this.Config.File.Version < 12)
-				this.SetupLegacy(false);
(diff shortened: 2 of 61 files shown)


sleepybnuuy (changes requested):
v3 migration comments:
- every time the plugin is disabled->enabled, the v3 migration dialog is re-fired on entering GPose afterwards
- toggles appear inverted during v3 setup; if `Enable Keybinds` and `UI buttons toggle windows` are both checked, with Toolbar unchecked, proceeding will initialize ktisis with Toolbar enabled and the other two settings disabled
- One of Keybinds/Toggle Windows must be enabled during V3 setup to proc the `Finish` button; if both are unchecked, the bottom-row button displays `Next` and sends to a null page regardless of Toolbar's check/uncheck state

what conditions need to be set up as far as config files are concerned to test the v2 migration portion? Simply running ktisis v2 then switching to this v3 build doesn't trigger the v2 migration on its own

Glorou:
Fixed the disable/enable issue.
I cant replicate two, will discuss with you shortly.
Fixed point 3, the idea was that it wouldnt show at all if you already had all the default settings already and you could just skip.

To activate the v2 migration, you cant have any v3 config, just rename your v3 conf to .old

sleepybnuuy (changes requested):
v2 migration comments:
- `Use Ktisis Toolbar` lists its default as On, but the checkbox is unticked by default. This means that proceeding without checking it initializes v3 config with toolbar turned off, despite the encouragement that it be set as the default
- needs at least an attempt at copying some of the existing keybinds over; there doesn't need to be a UI to set these inline, only a note that if keybinds are enabled during migration, that they'll be transferred from v2 where possible and otherwise use our defaults
- I'll take a writing/editing pass on the text once code changes are finalized

sleepybnuuy (commented):
keybinds and offsets now save correctly during v2 migration - still needs the following addressed (plus a writing pass from myself):
- offsets window sizing in Toolbar mode (and unreachable buttons in non-Toolbar mode)
- 2x modifier keys when importing KeyBinds from v2 (Ctrl+Ctrl+Q)

PR #314, opened 2026-06-16
IPC changes for TimelineAnimator

NoHideout:
Removed unused IPCs that were made for it
Added one new IPC to apply a pose directly
Changed Ktisis.SelectedBones to return all bones if a group is selected

### DIFF
--- a/Ktisis/Interop/Ipc/IpcProvider.cs
+++ b/Ktisis/Interop/Ipc/IpcProvider.cs
@@ -105,8 +105,9 @@ public class IpcProvider(ContextManager ctxManager, IDalamudPluginInterface dpi,
 
 			ret[actor.Actor.ObjectIndex] =
 				actor.Children.OfType<EntityPose>()
-					.SelectMany(x => x.Recurse())
+					.SelectMany(x => x.Recurse().Append(x))
 					.Where(s => s.IsSelected)
+					.SelectMany(s => s.Recurse().Append(s))
 					.OfType<BoneNode>()
 					.Select(s => s.Info.Name)
 					.ToHashSet();
--- a/Ktisis/Interop/Ipc/IpcProvider.cs
+++ b/Ktisis/Interop/Ipc/IpcProvider.cs
@@ -13,6 +13,7 @@ using Ktisis.Data.Files;
 using Ktisis.Data.Json;
 using Ktisis.Editor.Context;
 using Ktisis.Editor.Context.Types;
+using Ktisis.Editor.Posing;
 using Ktisis.Editor.Posing.Data;
 using Ktisis.Editor.Transforms;
 using Ktisis.Scene.Entities.Game;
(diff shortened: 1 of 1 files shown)


Ktisis/Interop/Ipc/IpcProvider.cs
```diff
@@ -120,165 +118,57 @@ private async Task<Dictionary<int, HashSet<string>>> SelectedBones() {
...
+				var bone = actor.Pose.FindBoneByName(kvp.Key);
+				if (bone == null) continue;
+				
+				var pose = bone.GetPose();
+				if (pose == null || pose->ModelPose.Data == null) continue;
+				
+				Matrix4x4.Decompose(kvp.Value, out var scale, out _, out _);
+				
+				var qsModel = pose->ModelPose.Data + bone.Info.BoneIndex;
+				qsModel->Scale = new hkVector4f { X = scale.X, Y = scale.Y, Z = scale.Z, W = 0f };
```
sleepybnuuy, line 168: does this need to be done in a separate loop after SyncModelSpace? if not, setting ModelPose scale could be done during the same kvp iteration as above

Ktisis/Interop/Ipc/IpcProvider.cs
```diff
@@ -120,165 +118,57 @@ private async Task<Dictionary<int, HashSet<string>>> SelectedBones() {
...
-		if (useWorldSpace) return bone.GetMatrix();
-
-		// ps relative
-		var model = bone.GetMatrixModel();
-		if (model == null) return null;
+	#region Absolute Posing IPC
 
-		var parentModel = GetParentBone(bone)?.GetMatrixModel() ?? Matrix4x4.Identity;
+	private ActorEntity? GetEntity(uint index) => ctxManager.Current?.Scene?.GetEntityForIndex(index);
+	private BoneNode? GetParentBone(BoneNode bone) => bone.Pose.Recurse().OfType<BoneNode>().FirstOrDefault(p => bone.IsBoneChildOf(p));
```
sleepybnuuy: function is unused with the obsolete ipc methods gone, you can remove this

Ktisis/Interop/Ipc/IpcProvider.cs
```diff
@@ -120,165 +118,57 @@ private async Task<Dictionary<int, HashSet<string>>> SelectedBones() {
...
-		var ctx = ctxManager.Current;
+	/// <summary>
+	/// Local Pos/Rot, Model Scale to circumvent Racial Offset
+	/// </summary>
+	private async Task<bool> ApplyAbsolutePoses(uint index, Dictionary<string, Matrix4x4> matrices) {
 		var actor = GetEntity(index);
-		var bone = actor?.Pose?.FindBoneByName(boneName);
-
-		if (ctx is null || bone is null) return false;
+		if (actor?.Pose == null || matrices.Count == 0) return false;
```
sleepybnuuy, line 130: for curiosity's sake - with the removal of the GetMatrix commands, where are your submitted matrices coming from now? are they formed from SavePose to get the raw bone transforms and modified by the timelineanimator, or pulled directly from havok instead of ipc?

sleepybnuuy (commented):
few notes before merge, largely looks good

NoHideout:
Just to have it noted down here as well;
Im getting the matrices directly from havok (Local Position and Rotation, Model Scale), 
Local Scale unfortunately already contains the Racial Scaling offset.
When ktisis would run SyncModelSpace on that scale, racial scaling would be multiplied once more on top of it.
I bypass that by applying Model Scale in the second loop directly. That works fine since scale does not need to be in local space.

I removed the unused GetParentBone function and added some caching to ApplyAbsolutePoses, that avoids doing a second decompose on the incoming matrices and in theory speeds the whole thing up (by fractions).

sleepybnuuy (approved):

PR #322, opened 2026-06-25
Karou's Cool Fixes

sleepybnuuy:


Ktisis/Data/Config/ConfigManager.cs
```diff
@@ -60,22 +60,21 @@ public void Load() {
...
-		}
+		if (cfg is null)
+			try {
+				cfg = this.CreateDefault();
+				this.GenerateDefaultPresets(cfg);
+				created = true;
+			} catch (Exception err) {
+				Ktisis.Log.Error($"Failed to create default configuration:\n{err}");
+				throw;
+			}
```
Fayti1703: Minor style nit: This `if` should have braces, since the body is multi-line:

```suggestion
		if (cfg is null) {
			try {
				cfg = this.CreateDefault();
				this.GenerateDefaultPresets(cfg);
				created = true;
			} catch (Exception err) {
				Ktisis.Log.Error($"Failed to create default configuration:\n{err}");
				throw;
			}
		}
```

PR #330, opened 2026-07-10
Speedy toolbar

sleepybnuuy:


Cazzar:
Thought on this - most of these movements are per-frame based, if we scaled them with a Lerp, we might be able to make it consistent, regardless of the framerate?

Might be overkill though, would need some pretty big global state.

sleepybnuuy:
shrinking feels pretty good, didnt run into any obvious issues with it. see discussion in internal-dev for other ideas on reducing toolbar size concerns

PR #336, opened 2026-07-19
will fix toolbar being unavailable when DalamudClassic is messed up

Glorou:
I want to fix this all properly at some point but this is a good start


Glorou:
Dont merge this without checking the style, it probably should be conditional on using the toolbar

sleepybnuuy (approved):
lgtm on my end with a few different ui styles, no clue how we're meant to replicate the original bug with fetching DalamudClassic but should work for affected users

PR #339, opened 2026-07-19
Attempted to fix import pose crash, and fixed viewport scaling with global scale

Glorou:


Ktisis/Interface/Types/EntityEditWindow.cs
```diff
@@ -33,9 +33,11 @@ public override void PreDraw() {
 	}
 
 	public virtual void SetTarget(T target) {
-		if (!target.IsValid)
+		try {
+			this.Target = target;
+		} catch {
 			throw new Exception("Attempted to set invalid target.");
-		this.Target = target;
+		}
```
sleepybnuuy, line 42: this would swallow an exception from setting the target and removes the validity check, I would suggest:

```cs
try {
    if (!target.IsValid)
        throw new Exception("Attempted to set invalid target.");
    this.Target = target;
} catch (Exception err) {
    Ktisis.Log.Error($"Attempting to set target raised exception:\n{err}");
}
```

Ktisis/Interface/Overlay/OverlayWindow.cs
```diff
@@ -55,7 +56,8 @@ public override void PreOpenCheck() {
 	}
 
 	public override void PreDraw() {
-		this.Size = ImGui.GetMainViewport().Size;
+		
+		this.Size = ImGui.GetMainViewport().Size / ImGuiHelpers.GlobalScale;
```
sleepybnuuy, line 60: Same should probably be done with Gizmo2D's SetNextWindowSize

sleepybnuuy (changes requested):

Ktisis/Interface/Overlay/OverlayWindow.cs
sleepybnuuy: @Glorou please do this in Gizmo2D too thx

sleepybnuuy (approved):
big thankies

PR #343, opened 2026-07-21
Create CharacterHooks for automatically adding new actors

Cazzar:
One consideration to add, is if this behavior should be a config setting, having it on by default would be a good thing if so.

This functionally restores the 0.2-like functionality of auto picking up characters.

Functionality as stands:

https://github.com/user-attachments/assets/2b24d618-89cd-40a5-95f0-c23444f2b343

> [!warning]
> This is going to require some heavy testing from all the methods of adding, to ensure we don't accidentally duplicate actors.


sleepybnuuy:
going to hold off on a merge for this today - runs into some odd behavior when active in conjunction with brio's Spawn Actor With Companion; _sometimes_ this will spit out extra empty entities into ktisis at the world origin, other times will not - at least getting this to consistently only add valid actors would be ideal imo, unless we just cant do anything to better manage these companion slots without more architecture work on our side

Cazzar:
> going to hold off on a merge for this today - runs into some odd behavior when active in conjunction with brio's Spawn Actor With Companion; _sometimes_ this will spit out extra empty entities into ktisis

I think this came down to the `this.AddActor(character, addCompanion)` where I was passing in to addCompanion, which at further thought seems pointless, as the companion would end up having the companion call the hook to.

This also caused a regression where 3 actors would get added if you had a companion slotted actor

Ktisis/Scene/Modules/Actors/ActorModule.cs
```diff
@@ -295,12 +289,105 @@ private unsafe void ControlGazeDetour(nint a1) {
...
+		if (gameObject is null || gameObject.ObjectIndex < 200) {
+			Ktisis.Log.Verbose("Unable to find gameobject, or below 200 for {0:X} ({1})", (nint)character, gameObject?.ObjectIndex);
+
+			return;
+		}
+
+		if (!gameObject.IsDrawing()) {
+			Ktisis.Log.Debug("Actor[{0:X} / {1}] Actor not drawing, not adding", (nint)character, gameObject.ObjectIndex);
+			return;
+		}
```
sleepybnuuy: should probably be removed; in testing, this prevented any brio actors from being created on the hook

Image

sleepybnuuy (changes requested):

sleepybnuuy (approved):
removed the isdrawing check per internal discussion

Glorou (approved):

PR #348, opened 2026-07-23
Multitarget fixes

sleepybnuuy:
from issue https://discord.com/channels/975894364020686878/1529681751419654184

- fixed naive same-name matching on the DrawTargetLabel for workspace's selection list causing an imgui break
- updated toolbar-exclusive behavior for ToggleEditorOnSelect; if the action was multi-mode, it now will not open the object editor with every indiv. addition to the group selection

### DIFF
--- a/Ktisis/Editor/Selection/SelectManager.cs
+++ b/Ktisis/Editor/Selection/SelectManager.cs
@@ -151,7 +151,7 @@ public class SelectManager : ISelectManager {
 	}
 
 	public void Select(SceneEntity entity) {
-		var removed = this.Selected.Remove(entity);
+		this.Selected.Remove(entity);
 		this.Selected.Add(entity);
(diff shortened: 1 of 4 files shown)


Ktisis/Editor/Selection/SelectManager.cs
```diff
@@ -151,9 +151,9 @@ public bool IsActorSelected(ActorEntity actor) {
 	}
 
 	public void Select(SceneEntity entity) {
-		this.Selected.Remove(entity);
+		var removed = this.Selected.Remove(entity);
```
Cazzar: We've added a var for the result of remove, but not used that, is this supposed to be used anywhere?

Cazzar (dismissed):

Cazzar (approved):

PR #351, opened 2026-07-23
Disable Camera Orbit Sections when Null Target

sleepybnuuy:
from issue https://discord.com/channels/975894364020686878/1004373141000290395/1529793529801474108

- instead of skipping draw for orbit/tracking, ImRaii.Disable them if we have a null orbit target for any reason
- this should prevent confusion and signal to the user to change targets in order to restore functionality; issue can occur when (for instance) an actor is deleted while being orbit-targeted
- some style/nitpick fixes to this file are included

this should be a decent half-step til we do more CameraWindow updates to add target-setting+camera switching inside the ui here


Glorou (approved):
Looks good to me, works in game and don't see any issues witht he code.

PR #352, opened 2026-07-23
Expression blending

chirpxiv:
I'll be travelling from the 24th until the 27th, so this is everything I was able to do before then. Some help would be appreciated to get this merged in that time if there's any glaring issues with it.

Draft left up to maintainer discretion on when it's ready to review/merge.

Notes:
- largely untested, especially on other races (non-miqo), please ensure it works well
  - especially test differences between midlanders/highlanders
- if there's better expressions to use please use them
- `FaceLibraryGenerator` should ideally be in its own project for easier maintenance, it doesn't depend on much
- generation can sometimes fail on the first run after game startup (this only matters to us)
- a clean way to 'reset' the face pose would be good; I was thinking of a reference slider that overwrites the current expression?
- there is a risk of the sliders falling out of sync with the current pose (ie. on pose import)
- sliders might need reordering somehow for UX/comfort (ie. jaw being lower, near similar mouth controls)
- missing chinese localisation strings

also fixes a propagation inaccuracy with transforms involving both rotation and translation (bd1c363)


sleepybnuuy:
merged in feature patch 1 to complete most of the todos, should be in a usable state as the first draft/release to iterate on wider feedback

Glorou (approved):

PR #354, opened 2026-07-24
add utility to regenerate default presets

sleepybnuuy:
reported via DM

- adds a `Regenerate Defaults` button to the PresetEditor to backfill any missing presets when used

image


Cazzar:
Just as a thought one of the issues I've had with this since I implemented, is our default groups aren't translatable in code right now - should we try adding that in?
Happy for this to be a follow-up too

sleepybnuuy:
> Just as a thought one of the issues I've had with this since I implemented, is our default groups aren't translatable in code right now - should we try adding that in? Happy for this to be a follow-up too

I'd like to skip this for now - localizing presets/precomputed data like that (e.g. Gobos list, Props csv) should probably be tackled in a future PR

Cazzar (approved):

PR #355, opened 2026-07-27
Automatically make it so the toolbar bottom cannot escape the bottom of the screen

Cazzar:
Initially reported here: https://discord.com/channels/975894364020686878/1004373141000290395/1531201909183217705

https://github.com/user-attachments/assets/1b81af87-bb12-4508-9130-834b8589e7b4

### DIFF
--- a/Ktisis/Interface/Windows/ToolbarModules/Workspace.cs
+++ b/Ktisis/Interface/Windows/ToolbarModules/Workspace.cs
@@ -34,7 +34,14 @@ public class Workspace : WorkspaceWindow  {
 		this._workspace.DrawCompact();
 
 		var botHeight = (UiBuilder.DefaultFontSizePx + (style.ItemSpacing.Y + style.ItemInnerSpacing.Y) * 2) * ImGuiHelpers.GlobalScale;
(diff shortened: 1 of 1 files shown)


sleepybnuuy (changes requested):
@Cazzar can we also do this for the auto-resizing Object Editor when `AutoResizeObjectEditor` is turned on?

Cazzar:
> can we also do this for the auto-resizing Object Editor when `AutoResizeObjectEditor` is turned on?

It might not be as easy, as we aren't arbitrarily setting a child container height.

One of the reasons for doing this is there was no user control to the max size in toolbar mode, whereas in the object editor you can collapse sections.

sleepybnuuy:
> One of the reasons for doing this is there was no user control to the max size in toolbar mode, whereas in the object editor you can collapse sections.

should be fine if it cant be included, i'll build and have a look at the toolbar portion today and we can get this out anyhow regardless

sleepybnuuy (approved):
fixes workspace height; other windows such as Object Editor will need more work to prevent them from being unable to draw their full contents when expanded. see attached - maximum height of any toolbar window is restricted by the height of the display viewport and with no scrollbars, its impossible to use everything at once
this should be a focus of toolbar improvement efforts/reworks

Image

PR #358, opened 2026-07-27
Import pose fixes

sleepybnuuy:
- prevent PreviewNode from hiding when game UI is toggled off
- add ApplyOnSelect similar to NPC imports to Pose Imports; this depends on an event from FileSelect that, when seen in PoseImportDialog, will automatically apply the pose without the user needing to hit the button
- disabled the Apply button if the user is not in posemode

### DIFF
--- a/Ktisis/Interface/KTK/PreviewNode.cs
+++ b/Ktisis/Interface/KTK/PreviewNode.cs
@@ -32,6 +32,7 @@ namespace Ktisis.Interface.KTK;
 public unsafe class PreviewNode : OverlayNode {
 	public override OverlayLayer OverlayLayer => OverlayLayer.BehindUserInterface;
 	public override bool HideWithNativeUi => false;
+	public override bool HideWithUiToggled => false;
 	public override bool IsVisible { get; set; } = false;
 	protected override void OnUpdate() { }
 
--- a/Ktisis/Interface/Windows/Import/PoseImportDialog.cs
+++ b/Ktisis/Interface/Windows/Import/PoseImportDialog.cs
@@ -33,11 +33,21 @@ public class PoseImportDialog : EntityEditWindow<ActorEntity> {
 	) {
 		this._select = select;
 		select.OnOpenDialog = this.OnFileDialogOpen;
+		select.OnSelectedFile = this.OnFileSelected;
 	}
 
 	private void OnFileDialogOpen(FileSelect<PoseFile> sender) {
 		this.Context.Scene.Overlay.ToggleCharaViewTexture(this.Context, this.Target);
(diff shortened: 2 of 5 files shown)


Ktisis/Interface/Windows/Import/PoseImportDialog.cs
```diff
@@ -33,13 +33,23 @@ FileSelect<PoseFile> select
 	) {
 		this._select = select;
 		select.OnOpenDialog = this.OnFileDialogOpen;
+		select.OnSelectedFile = this.OnFileSelected;
```
Cazzar, line 36: Looking at this, one thing that comes to mind - are we possibly leaking a `this` reference back into the delegate? Should we explicitly handle this and remove?

Ktisis/Interface/KTK/PreviewNode.cs
```diff
@@ -32,6 +32,7 @@ namespace Ktisis.Interface.KTK;
 public unsafe class PreviewNode : OverlayNode {
 	public override OverlayLayer OverlayLayer => OverlayLayer.BehindUserInterface;
 	public override bool HideWithNativeUi => false;
+	public override bool HideWithUiToggled => false;
```
Cazzar, line 35: One side effect of this is if you leave draggable enabled and toggle UI - the outline is still visible, do we want to toggle that off too?

Cazzar (changes requested):
No major isuses - a couple of thoughts/comments on two of the things.

UX wise, I was looking for the error mesage around "you need to be in pose mode" much earlier in the pipeline - I would possibly think about having the warning block all actions in the UI with that message instead?

Ktisis/Interface/KTK/PreviewNode.cs
sleepybnuuy: previewnode should never be draggable, and this flag specifically makes the node _not_ hide when the UI is toggled. all our other nodes have this flag set too; you might be confusing it with normal visibility+draggable? which i'm fine leaving in place

sleepybnuuy:
> UX wise, I was looking for the error mesage around "you need to be in pose mode" much earlier in the pipeline - I would possibly think about having the warning block all actions in the UI with that message instead?

folks will still want to be able to browse poses with posemode off, especially with the previewnode being added, so i'm comfortable leaving the rest of the window interactable and only stopping the application with this tooltip. if it's preferred, we could move to a less subtle warning label above the file selector, but i'd still like to keep the rest of the window un-disabled

Cazzar:
> Folks will still want to be able to browse poses with posemode off, especially with the previewnode being added

But the preview node requires pose mode to be on to work?

Cazzar (approved):

PR #364, opened 2026-07-28
Tweaked Suggestions

sleepybnuuy:
- sort Collections/Designs alphabetically in IPC tab
- added keybind to Toggle Siblink
- allow bulk deletion from workspace when multiple IDeletable entities are selected
- improved Pose Window actor targeting logic to more reliably stay/find focus on a particular actor
- added warning text in the popup for Show All Bones
- fixed Gaze Link button conflicting with Pose Parenting button


Glorou (dismissed):
Code is good, compiles runs and functions as intended, worth discussing: Bulk delete might warrant a confirmation dialogue

PR #367, opened 2026-07-31
Syncing game and draw obj

Glorou:
This just makes sure the draw object and the game object are always in sync, the camera bases its rotation off of where the gameobj is, not the draw, so there should no longer be a need to offset the camera.


Ktisis/Scene/Entities/Game/ActorEntity.cs
```diff
@@ -368,4 +369,17 @@ public unsafe uint GetActorGazeTarget() {
...
 	}
+	
+	
+	public unsafe override void SetTransform(Transform trans) {
+		var drawPtr = this.GetObject();
+		var gameObjPtr = this.Character;
+		if (drawPtr == null || gameObjPtr == null) return;
+		drawPtr->Position = trans.Position;
+		drawPtr->Rotation = trans.Rotation;
+		drawPtr->Scale = trans.Scale;
```
sleepybnuuy: this would need to use base.SetTransform so it adheres to CharaEntity's attach handling and to cut down on code duplication

sleepybnuuy (changes requested):
this shouldn't ship until we have a solution for camera jank when manipulating the gpose-focused actor. options include:
- not updating game object position if we're targeting the actor being moved
- only updating game object position when we change targets; this would mean you could move the targeted actor, switch to another, then when you switch back the perspective will shift to the new position
- when targeted actor is moved - set a dirty/needs-transform-update flag in SetTransform instead of immediately performing the update to game object's position. in Update(), if the actor has the dirty flag and isn't the camera target, update the gameobject position _(this would be my preference)_

sleepybnuuy (dismissed):
good from my testing but will need another reviewer before merge

Glorou:
@sleepybnuuy The way this works feels a bit counter-intuitive, but I think it would be good enough if we swapped the offset camera button functionality for this in the camera editor, since that's what users know to click.

chirpxiv (approved):

PR #372, opened 2026-08-04
Fix missing language during V2->V4 migration

sleepybnuuy:
- forces fallback language to always be loaded, so when we have no config (as during v2->v4 migratorwindow), we can still draw localized english
- reformats the file bc spacing and vars were bothering me


sleepybnuuy (commented):
fayti change lgtm

chirpxiv (approved):

PR #374, opened 2026-08-08
ObjectTable / Init Hooks fixes

sleepybnuuy:
- adds ExpectedIndex to ActorModule: this indicates the currently-anticipated spawn on the object table, and is checked by ActorModule.Add to make sure we don't add an actor via initialize when we've pressed the button to create them ourselves.
  - this should fix issues where a duplicate actor using the PC's name could be spawned when manually pressing Add Actor or when loading a ktscene
- dynamically staggers ActorModule.Add runs by the count of encountered characters we've yet to process
  - this should aid the above issue (by slowing the possible race condition that causes Module and Spawner to collide)
  - this should also fix cases where Actor 202 ends up in the Workspace above Actor 201 - this was occurring when 200, 201, and 202 would be fired sequentially, but the framework tick they were all scheduled on would pick them up in different orders. with the stagger in place, 202 will now be guaranteed to fire one tick later than 201 if they were enqueued on the same frame, ensuring that they're first in the scenetree's ienumerable
  - this will matter less when we finish Folders work for reordering/dynamically sorting the scenetree
- un-obsoletes CalculateNextAvailableIndex

### DIFF
--- a/Ktisis/Scene/Modules/Actors/ActorModule.cs
+++ b/Ktisis/Scene/Modules/Actors/ActorModule.cs
@@ -296,16 +296,16 @@ public class ActorModule : SceneModule {
 		Ktisis.Log.Verbose("[Initialize] New Character? {0:X}", (nint) thisPtr);
 		
 		try {
-			_initializeHook.OriginalDisposeSafe(thisPtr);
+			this._initializeHook.OriginalDisposeSafe(thisPtr);
 		} catch (Exception e) {
 			Ktisis.Log.Error(e, "Error on Initialize");
 		}
 		
 		if (!this.CheckValid()) return;
 		
-		_framework.RunOnTick(() => {
+		this._framework.RunOnTick(() => {
 			this.Add(thisPtr);
-		}, delayTicks: 1); //delayed to allow internal code to handle 
+		}, delayTicks: 5); // delayed to allow internal code to handle, wait a bunch of ticks in case of chicanery
 	}
 	
 	private unsafe GameObject* Destructor(Character* thisPtr, byte freeFlags) {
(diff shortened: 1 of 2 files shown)


Ktisis/Scene/Modules/Actors/ActorModule.cs
```diff
@@ -355,26 +361,28 @@ private unsafe void Remove(Character* character) {
 
 	private unsafe void Add(Character* character) {
 		var gameObject = this._actors.GetAddress((nint)character);
-		if (gameObject is null || gameObject.ObjectIndex < 200) {
-			Ktisis.Log.Verbose("Unable to find gameobject, or below 200 for {0:X} ({1})", (nint)character, gameObject?.ObjectIndex);
-
+		if (gameObject is null) {
+			Ktisis.Log.Verbose($"{(nint)character:X} - GameObject is null");
+			return;
+		} if (gameObject.ObjectIndex <= 200) {
```
Cazzar: nitpick on code style

```suggestion
		} 
		
		if (gameObject.ObjectIndex <= 200) {
```

or if you want same line


```suggestion
		} else if (gameObject.ObjectIndex <= 200) {
```

I know functionally they're the same but at a glance it looks like an error.

Ktisis/Scene/Modules/Actors/ActorModule.cs
```diff
@@ -293,23 +296,26 @@ private unsafe void ControlGazeDetour(nint a1) {
...
 		if (!this.CheckValid()) return;
-		
-		_framework.RunOnTick(() => {
+
+		this._enqueuedChars.Add(addr);
+		this._framework.RunOnTick(() => {
 			this.Add(thisPtr);
-		}, delayTicks: 1); //delayed to allow internal code to handle 
+			this._enqueuedChars.Remove(addr);
+		}, delayTicks: 1 + this._enqueuedChars.Count); // delayed to allow internal code to handle - stagger tick timings so that these can resolve in order of appearance
```
sudojunior, line 315: `ActorModule._enqueuedChars` does not appear to be checked for potential duplicates. _Unsure how much this matters with it being part of `InitializeHook`_

image

sudojunior (changes requested):
HashSet query on blocking duplicate calls.

_`ObjectIndex` in unmodified code was thought to be double checking what has already been verified (including with additional code), but does still seem to cover specific scenarios which all checks remain required._

Ktisis/Scene/Modules/Actors/ActorModule.cs
sleepybnuuy: hashsets are inherently unique, doing a .Add with an element already in the set returns False
similarly, .Remove returns False if the item was _not_ present in the set
this allows us to blind add/remove the unique nints of each pointer we try to process

sleepybnuuy: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.add?view=net-10.0#system-collections-generic-hashset-1-add(-0)

sudojunior (approved):

PR #375, opened 2026-08-10
Add side swap for PreviewNode.cs

### LINKED ISSUE #379: Preview Actor is offset relative to Windowed Mode position
When `Open Pose File` dialog is open, the preview actor will offset itself an lateral and vertical distance relative to the position of the game - when in windowed mode.

https://github.com/user-attachments/assets/6091fc29-28c7-4016-8552-9e9aca357c40

Discovered while testing and making changes for #375.

sudojunior:
Model preview will move to the left of FileDialog if it does not have enough space to render on the right.

https://github.com/user-attachments/assets/49b8831a-21f2-48e6-80f9-8955c159e0eb

### DIFF
--- a/Ktisis/Interface/KTK/PreviewNode.cs
+++ b/Ktisis/Interface/KTK/PreviewNode.cs
@@ -166,7 +166,7 @@ public unsafe class PreviewNode : OverlayNode {
 		}
 
 		this.IsVisible = true;
-		var offset = this._fileWindow.Pos.X + this._fileWindow.Size.X + this.Border.Size.X > ImGui.GetIO().DisplaySize.X
+		var offset = this._fileWindow.Pos.X + this._fileWindow.Size.X + this.Border.Size.X > ImGui.GetMainViewport().Size.X
 			? -this.Border.Size.X : this._fileWindow.Size.X;
 		this.Position = new Vector2(this._fileWindow.Pos.X + offset, this._fileWindow.Pos.Y);
 
--- a/Ktisis/Interface/KTK/PreviewNode.cs
+++ b/Ktisis/Interface/KTK/PreviewNode.cs
@@ -166,7 +166,9 @@ public unsafe class PreviewNode : OverlayNode {
 		}
 
 		this.IsVisible = true;
-		this.Position = new Vector2(this._fileWindow.Pos.X + this._fileWindow.Size.X, this._fileWindow.Pos.Y);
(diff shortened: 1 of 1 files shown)


Ktisis/Interface/KTK/PreviewNode.cs
```diff
@@ -166,7 +166,9 @@ private void OnFramework(IFramework framework) {
 		}
 
 		this.IsVisible = true;
-		this.Position = new Vector2(this._fileWindow.Pos.X + this._fileWindow.Size.X, this._fileWindow.Pos.Y);
+		var offset = this._fileWindow.Pos.X + this._fileWindow.Size.X + this.Border.Size.X > ImGui.GetIO().DisplaySize.X
```
sleepybnuuy: use ImGui.GetMainViewport().Size instead of DisplaySize, since the user can be in Windowed mode which will give back a smaller resolution. also examine whether Viewport size needs to be multiplied by ImGuiHelpers.GlobalScale, which we do in OverlayWindow - test this by changing global scale in dalamud's native ui settings after using getmainviewport

sleepybnuuy (changes requested):

Ktisis/Interface/KTK/PreviewNode.cs
```diff
@@ -166,7 +166,9 @@ private void OnFramework(IFramework framework) {
 		}
 
 		this.IsVisible = true;
-		this.Position = new Vector2(this._fileWindow.Pos.X + this._fileWindow.Size.X, this._fileWindow.Pos.Y);
+		var offset = this._fileWindow.Pos.X + this._fileWindow.Size.X + this.Border.Size.X > ImGui.GetMainViewport().Size.X
+			? -this.Border.Size.X : this._fileWindow.Size.X;
+		this.Position = new Vector2(this._fileWindow.Pos.X + offset, this._fileWindow.Pos.Y) - ImGui.GetMainViewport().Pos;
```
sudojunior, line 171: Not strictly necessary, but common reference of viewport.
```suggestion
		var viewport = ImGui.GetMainViewport();
		var offset = this._fileWindow.Pos.X + this._fileWindow.Size.X + this.Border.Size.X > viewport.Size.X
			? -this.Border.Size.X : this._fileWindow.Size.X;
		this.Position = new Vector2(this._fileWindow.Pos.X + offset, this._fileWindow.Pos.Y) - viewport.Pos;
```

Ktisis/Interface/KTK/PreviewNode.cs
sudojunior: Resolved with f6e37aa

sleepybnuuy (approved):

PR #376, opened 2026-08-10
Disallow Actor #441 from showing in workspace

sudojunior:
* Conflicts with PreviewNode when FileDialog is open


sudojunior:
Superseded by #377

PR #384, opened 2026-08-24
Disable actor renaming if incognito is active

sudojunior:
* Requires ktisis-tools/GLib@ba9205541292f1667a4f0fb0d3f902f725a3890f via ktisis-tools/GLib#7

image

*While incognito is active. Unsure if this is the intended solution, but could also add a note or tooltip somewhere here about incognito being active.*


sleepybnuuy (approved):
dig it

sleepybnuuy:
will need GLib latest commit e17c9609af15a24e9c9ff03929175a416834aac5 as part of release

PR #390, opened 2026-09-14
Suppresses Toast notifications while in GPose

Glorou:


sleepybnuuy:
needs to be tweaked to actually allow on/off toggling

